Issue #5, opened 2016-11-30
Fix init file


jonDel:
Fix package init file by removing dummy helloworld function and including the proper import statements, making calls to functions/classes more straightforward, without the need to refer to the submodules.
Ex:

```
#python3
>>>from serenata_toolbox import CEAPDataset
>>>from serenata_toolbox fimport fetch_latest_backup
>>>from serenata_toolbox import Reimbursements

```

PR #6, opened 2016-11-30
Fixing datasciencebr/serenata-toolbox#5 and datasciencebr/serenata-toolbox#7 issues

jonDel:
Pull content:
- Fixing init file;
- Removing unused imports and local variables.
 - Fixing script xml2csv.py location and reference in setup.py
References:  
datasciencebr/serenata-toolbox#5
datasciencebr/serenata-toolbox#7


cuducos:
Isn't it more coherent to adapt `xml2csv.py` to a module/class and use it within other scripts (instead of running it as a script)?

jonDel:
I did not realize that the script was called from inside a python module. @cuducos , indeed that call seems a bit strange. I thought that script was meant to be used as a tool to be invoked from command line. If that was not the original intention, and it was meant only to be used inside that module, I think is better refactor the script into a class/function inside that module.

jonDel:
By the way, I can do that. No problems. @Irio , what do you think?

Irio:
@jonDel That would be ideal. Go for it.

cuducos:
That'd be awesome, @jonDel. Many thanks. Also you've removed the `hello_world` but the docs at `README.md` are still referring to it. Either we keep that dummy function as a way to test if the module was installed and loaded into the `sys.path` or we have to update the docs. Does that make sense?

jonDel:
I'll update that either.

jonDel:
@Irio, the requested changes were made.

jonDel:
It is ok to be merged?

PR #16, opened 2016-12-12
Fetch speeches script

fgrehm:
_Follow up of https://github.com/datasciencebr/serenata-de-amor/issues/137#issuecomment-266220208_

- Initial hacking: https://gist.github.com/fgrehm/52a95cf3ccc95eaa0eead22589c8303d
- Playing with the data: https://gist.github.com/fgrehm/52a95cf3ccc95eaa0eead22589c8303d

As I mentioned on that issue, I'm new to Python so feel free to throw as many 🍅 as you want, I'd be really happy to learn more 😄 

/cc @cuducos @Irio 

-----------

- Initial analysis: https://gist.github.com/fgrehm/01ec6027b50a8c686e9da84ee1059ba3
- Datasets with 2015-2016 speeches
  - [2016-12-15-speeches.xz.zip](https://github.com/datasciencebr/serenata-toolbox/files/666302/2016-12-15-speeches.xz.zip) (the one I used on the notebook above)
  - [2016-12-21-speeches.xz.zip](https://github.com/datasciencebr/serenata-toolbox/files/666318/2016-12-21-speeches.xz.zip) (just collected this one today)

### DIFF
--- a/serenata_toolbox/speeches.py
+++ b/serenata_toolbox/speeches.py
@@ -6,18 +6,6 @@ import urllib
 
 from datetime import datetime
 
-"""
-Usage:
-
-    range_start = '01/11/2016'
-    range_end = '30/11/2016'
-    output = 'data/2016-11-speeches.xz'
-
-    speeches = Speeches()
-    df = speeches.fetch(range_start, range_end)
-    speeches.write_file(output, df)
-"""
-
 class Speeches:
 
     URL = (
@@ -114,3 +102,13 @@ class Speeches:
     @staticmethod
     def __extract_text(node, xpath):
         return node.find(xpath).text.strip()
+
+def fetch_speeches(output, range_start, range_end):
+    """
+    :param output: (str) directory in which the output file will be saved
+    :param range_start: (str) date in the format dd/mm/yyyy
+    :param range_end: (str) date in the format dd/mm/yyyy
+    """
+    speeches = Speeches()
+    df = speeches.fetch(range_start, range_end)
+    speeches.write_file(output, df)
--- a/serenata_toolbox/speeches.py
+++ b/serenata_toolbox/speeches.py
@@ -2,8 +2,9 @@ import os
 import pandas as pd
 import numpy as np
 import xml.etree.ElementTree as ET
-
 import urllib
+
+from datetime import datetime
 
 class Speeches:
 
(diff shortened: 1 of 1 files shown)


serenata_toolbox/speeches.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def fetch(self, range_start, range_end):
+        range = {'dataIni': range_start, 'dataFim': range_end}
+        url = self.URL.format(**range)
+        file=urllib.request.urlopen(url)
+
+        t = ET.ElementTree(file=file)
+        records = []
+        root = t.getroot()
+        for i, session in enumerate(root):
```
cuducos: I might have missed it, but it looks like that you're not using the `i` (the index returned by the `enumerate`). Can we use just `for session in root:`?

serenata_toolbox/speeches.py
```diff
@@ -0,0 +1,88 @@
...
+import pandas as pd
+import numpy as np
+import xml.etree.ElementTree as ET
+import urllib
+
+from datetime import datetime
+
+class Speeches:
+
+    URL = 'http://www.camara.leg.br/SitCamaraWS/SessoesReunioes.asmx/ListarDiscursosPlenario?dataIni={dataIni}&dataFim={dataFim}&codigoSessao=&parteNomeParlamentar=&siglaPartido=&siglaUF='
```
cuducos: For readability sake you can use [a trick](https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation) when the string is too long: parenthesis without commas (otherwise you break it into a tuple). This like would become something like (and the value of the string would be exactly the same)

```python
    URL = (
        'http://www.camara.leg.br/SitCamaraWS/SessoesReunioes.asmx/ListarDiscursosPlenario'
        '?dataIni={dataIni}'
        '&dataFim={dataFim}'
        '&codigoSessao=&parteNomeParlamentar=&siglaPartido=&siglaUF='
    )
```

serenata_toolbox/speeches.py
```diff
@@ -0,0 +1,88 @@
...
+                        phase_code,
+                        phase_desc,
+                        speech_speaker_num,
+                        speech_speaker_name,
+                        speech_speaker_party,
+                        speech_speaker_state,
+                        speech_started_at,
+                        speech_room_num,
+                        speech_insertion_num
+                    ])
```
cuducos: Actually I would wrap all the code between lines 29 and 56 in a new method called, for example `def serialize(session)`. 

Then I'd change `records = []` (line 26) to a generator: `records = map(self.serialize, root)`

serenata_toolbox/speeches.py
```diff
@@ -0,0 +1,88 @@
...
+        df.to_csv(filepath, **self.CSV_PARAMS)
+
+        print('Done.')
+
+# TODO: Read date range and output path from user input
+range_start = '01/11/2016'
+range_end = '30/11/2016'
+output = '.'
+
+if __name__ == '__main__':
```
cuducos: I don't think this bit (lines 85-88) is need in the repo, but it's damn useful for dev, I know ; ) But before merging it would be a good idea to remove that and document how to use with [docstrings](https://www.python.org/dev/peps/pep-0257/) — I know almost anything in this repo uses _docstrings_ but we should be using and it's on my task lisk ; )

serenata_toolbox/speeches.py
cuducos: In other cases I'd agree, but not here. The thing is that if the script doesn't need the `i` at all, just get rid of the function that was _forcing_ its creation (`enumerate`). Otherwise in one line you're using a function to create something (`enumerate`) to throw it away (`_`)…

fgrehm:
@jonDel @cuducos how does https://github.com/datasciencebr/serenata-toolbox/pull/16/commits/dbec7429de3afa427fcd6172e83442ccb7c7023f look to you guys?

I'll try to get the `__main__` removal + docstrings stuff later after work

fgrehm:
Hey guys, I'm done with the script. LMK if it needs any additional documentation, here's some info about the data and how we can use it: https://gist.github.com/fgrehm/bb0e1f6fef55082074d9a0258cf45391

LMK if you believe I should send a PR to the serenata-de-amor project with the notebook so that other people can start playing with the data

serenata_toolbox/speeches.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+Usage:
+
+    range_start = '01/11/2016'
+    range_end = '30/11/2016'
+    output = 'data/2016-11-speeches.xz'
+
+    speeches = Speeches()
+    df = speeches.fetch(range_start, range_end)
+    speeches.write_file(output, df)
```
cuducos: To be easily used in other script ([example](https://github.com/datasciencebr/rosie/blob/master/rosie/main.py#L18-L25)) you could wrap this in a function, something like:

```python
def fetch_speeches(output, start, end):
    """
    :param output: (str) directory in which the output file will be saved
    :param start: (str) date in the formar dd/mm/yyyy
    :param end: (str) date in the formar dd/mm/yyyy
    """
    speeches = Speeches()
    df = speeches.fetch(start, end)
    speeches.write_file(output, df)
```

This way people can used it like this:

```python
from serenata_toolbox.speeches import fetch_speeches
fetch_speeches('01/01/1970', '31/12/2016')  # done ; )
```

fgrehm:
@cuducos how about https://github.com/datasciencebr/serenata-toolbox/pull/16/commits/de5507924749157904f10181dad0ace7003a6570? :)

cuducos:
LGTM, many thanks for that, @fgrehm!
@Irio any extra comment data science wise? May I merge?

Irio:
Excellent, @fgrehm. Ready to merge for me. Have you collected the dataset? If so, could you attach it here so it can be added in our S3 bucket?

fgrehm:
Yup, I'll send it to you guys tomorrow along with some other new goodies 😄

fgrehm:
@Irio done! link to datasets are in the issue description

Irio:
Datasets added to the list with https://github.com/datasciencebr/serenata-toolbox/commit/6126436ba446ed56e3b291d73e61519d7c897a75.

@fgrehm If you open a pull request to datasciencebr/serenata-de-amor with the second notebook it can be merged without thinking twice. 👏

PR #22, opened 2017-01-24
Fix parser (case insensitive)

cuducos:
The new XML delivered has some inconsistencies: sometimes it used `<ideDocumento>…</ideDocumento>`, and others `<idedocumento>…</idedocumento>` This PR makes our parser case insensitive so it work with both tags.


Irio:
You must probably change attribute names in the `__translate_file` function. [ceap_dataset.py#L66](https://github.com/datasciencebr/serenata-toolbox/blob/3161805ba146b5392b666a6d2b5f6320fd676386/serenata_toolbox/ceap_dataset.py#L66)

PR #30, opened 2017-02-20
Create Dockerfile

Irio:


lipemorais:
The purpose of this just for production environment?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM continuumio/anaconda3:latest
```
cuducos: I would not use Anaconda Python here. IMHO `pip` installable packages should be based on the default Python installation (and specify the requirements in `setup.py`).

cuducos:
@lipemorais I think both guesses are valid and not exclusive. Am I wrong?

Dockerfile
lipemorais: In order to have a cleaner environment I would use [python:3](https://hub.docker.com/_/python/) image.

lipemorais:
@cuducos, it's not exclusive. For now looks that the Dockerfile is aiming emulate "production" environment, since it's necessary build the image every time that a change is made.

To make development easier could be used a shared folder with source code in this Dockerfile or a docker-compose.yml doing the same and keep this Dockerfile.

cuducos:
Got it. IMHO the `Dockerfile` in the case of a `pip` installable is more useful for tests — if we need to prioritize, that would be my first guess…

lipemorais:
Nice! Let me see if I got you too.

This is the kind of "test" that will help us keep the installation process always working, like running once a day or before release just to assure that everything are well.

> — if we need to prioritize, that would be my first guess…
I didn't get what you mean with it. :/

cuducos:
Sorry about that. Rephrasing it:

If we need to prioritize between _developing_ and _testing the installation_ as the focus of this `Dockerfile`, _testing the installation_ would be my first choice ; )

Irio:
Just made a new version of this Dockerfile, now with Alpine.

Purpose? Make development with toolbox easier, specially when working with other Serenata projects.

anaschwendler:
The objective of this PR is to facilitate the way to run `serenata-toolbox` using `Makefile`.
By now, we can see that we can run `make` and the option.

To test this PR I:
* First, I checked out to @Irio's branch:
```console
git checkout -b irio-docker origin/irio-docker
```
* Then I tested the `make` tasks:
```console
make bootstrap
```
This bootstrap the enviroment.

image

and then:
```console
make test
```
and ran the test suite.

image

And it LGTM ;)

Makefile
```diff
@@ -0,0 +1,9 @@
+default: test
+
+bootstrap:
+	python setup.py develop
+
+setup:
```
lipemorais: This make task is empty. Could it be removed or there a especial motive to have it here?

lipemorais:
I know that there is some recommendations for toolbox be used on Windows. Is it true for development too? If yes, `make` do not work on Windows environment as it do on Unix like ones. So this `Makefile` will not help who contribute from Windows.

lipemorais:
**cc** @Irio @anaschwendler

cuducos:
Considering my review and @lipemorais's feedback: is it possible to rollback this PR?

PR #33, opened 2017-03-28
Updates reimbursements file

### LINKED ISSUE #32: Update reimbursements file
We need a new reimbursements file on Amazon

jtemporal:
Closes #32

### DIFF
--- a/serenata_toolbox/datasets.py
+++ b/serenata_toolbox/datasets.py
@@ -27,7 +27,6 @@ def fetch_latest_backup(destination_path,
         '2016-11-19-current-year.xz',
         '2016-11-19-last-year.xz',
         '2016-11-19-previous-years.xz',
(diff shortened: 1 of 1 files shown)


serenata_toolbox/datasets.py
```diff
@@ -27,7 +27,6 @@ def fetch_latest_backup(destination_path,
         '2016-11-19-current-year.xz',
         '2016-11-19-last-year.xz',
         '2016-11-19-previous-years.xz',
-        '2016-11-19-reimbursements.xz',
```
Irio: Why have you removed this reimbursements file?

jtemporal: As I was discussing this with @cuducos, we came to agree that this method should download only the latest version of said data, even though the old file is still on the server and can be downloaded using the `fetch` method.

PR #35, opened 2017-03-29
Refactor datasets module: enhancements in terms of UX & async

cuducos:
This PR refactors `datasets` modules without changing it's public API — `serenata_toolbox.datasets.fetch(…)` and `serenata_toolbox.datasets.fetch_latest_backup(…)` still works as it used to work before.

There are two improvements: use of async methods to download multiple files in parallel and a better UX (a progress bar with the total size of the download and the progress).

untitled 3

### DIFF
--- a/serenata_toolbox/datasets.py
+++ b/serenata_toolbox/datasets.py
@@ -58,7 +58,6 @@ class Downloader:
 
         loop = asyncio.get_event_loop()
         loop.run_until_complete(self.main(loop, files))
-        loop.close()
 
     async def main(self, loop, files):
         if len(files) == 1:
--- a/serenata_toolbox/datasets.py
+++ b/serenata_toolbox/datasets.py
@@ -1,22 +1,18 @@
+import asyncio
 import os
-from urllib.request import urlretrieve
+
+import aiofiles
+import aiohttp
+from tqdm import tqdm
+
+AWS_BUCKET = 'serenata-de-amor-data'
+AWS_REGION = 's3-sa-east-1'
+MAX_REQUESTS = 4
 
 
-def fetch(filename, destination_path,
-          aws_bucket='serenata-de-amor-data',
-          aws_region='s3-sa-east-1'):
-    url = 'https://{}.amazonaws.com/{}/{}'.format(aws_region,
-                                                  aws_bucket,
-                                                  filename)
-    filepath = os.path.join(destination_path, filename)
-    if not os.path.exists(filepath):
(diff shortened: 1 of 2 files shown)


jtemporal:
from serenata_toolbox import datasets

jtemporal:
Hey @cuducos great refactor! Thanks a lot for the version bump ;)
So I fired up my ipython and ran the following
```python
from serenata_toolbox import datasets
datasets.fetch_latest_backup('/Users/temporal/toolbox-data/')
```
Which worked perfectly, but I tried to run in the same ipython session `datasets.fetch('2017-03-15-reimbursements.xz', '/Users/temporal/2-toolbox-data/')` and got the following error:
```console
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-5eabdb6d99be> in <module>()
----> 1 datasets.fetch('2017-03-15-reimbursements.xz', '/Users/temporal/2-toolbox-data/')

/Users/temporal/src/dsbr/serenata/serenata-toolbox/serenata_toolbox/datasets.py in fetch(filename, destination_path, aws_bucket, aws_region)
    113 def fetch(filename, destination_path, aws_bucket=None, aws_region=None):
    114     downloader = Downloader(destination_path, aws_bucket, aws_region)
--> 115     return downloader.download([filename])
    116 
    117 

/Users/temporal/src/dsbr/serenata/serenata-toolbox/serenata_toolbox/datasets.py in download(self, files)
     58 
     59         loop = asyncio.get_event_loop()
---> 60         loop.run_until_complete(self.main(loop, files))
     61         loop.close()
     62 

/Users/temporal/anaconda3/envs/toolbox/lib/python3.6/asyncio/base_events.py in run_until_complete(self, future)
    441         Return the Future's result, or raise its exception.
    442         """
--> 443         self._check_closed()
    444 
    445         new_task = not futures.isfuture(future)

/Users/temporal/anaconda3/envs/toolbox/lib/python3.6/asyncio/base_events.py in _check_closed(self)
    355     def _check_closed(self):
    356         if self._closed:
--> 357             raise RuntimeError('Event loop is closed')
    358 
    359     def _asyncgen_finalizer_hook(self, agen):

RuntimeError: Event loop is closed
```
I highly doubt that someone will run both `fetch_latest_backup()` and `fetch()` in the same session, but it would be nice not to have this outcome, do you think we can do something about this? maybe add another `event loop` to separate both methods?

cuducos:
Makes sense. The event loop was closed — fix in 2322c08

PR #36, opened 2017-04-05
Changing reimbursement data collection to conform new version

luizcavalcanti:
The brazilian congress made available a new version of the reimbursement
dataset, which now can be downloaded already in csv format and separated
by year. This somewhat simplifies the current fetching routine, but it
also demands a lot of small changes in the way the dataset is aquired,
translated, loaded and merged.

I this rather large commit, the unit tests were also modified to conform
to the new way of fetching, and the "convert to csv" test was removed,
since this operation is not needed anymore. The convert_to_csv() mehtod
from the CEAPDataset class was kept, though, so we don't break rosie
until it removes the call to this method.

Another quite drastic change was made in the test cases' names. A number
was added to them, so we can 'make sure' a proper order is followed
during integration tests. I strongly believe there is a better way of
doing this. I can work on it in a near-future commit.

Signed-off-by: Luiz Carlos Cavalcanti <cavalcanti.luiz@gmail.com>


luizcavalcanti:
This change is part of something I was talking to @cuducos on the project's telegram channel. The idea, as explained in the commit message, is to conform to the new data format for reimbursements. Besides the obvious XML->CSV change, there were little annoyances that had to be corrected to make it work. The data is not perfect, there are entries that have more columns than defined originally (specially in 2009 data), and I have no idea if congress IT people will correct it someday, so a panda  parameters were used to make the data loading process more lenient.

cuducos:
Ok… I just took a look (haven't ran any code yet, just read through it). Right now I have two main concerns:

* Can you document (here i the PR) the difference in terms of data structure? I mean, if they are just missing fields, I think there's no problem, but earlier we already have to deal with different subquota numbers/names etc. This should be done very carefully because we do use some of this data for the classifiers.
* After reading your considerations on tests I started to question the structure of the test case: IMHO it's actually an _integration test_ (it tests bunches of things at once, reading/writing to the filesystem, for example) but they are structured as _unit tests_ (test x, then test y, then test z). It  makes no sense. I would like to suggest a single test method telling a story, a proper integration test (example below).

<details>
Just drafting, but anyway…

```python
class TestFetchreimbursements(TestCase):

    def test_fetch_reimbursements(self):

        # let's create a temp dir for the sake of this tests
        …

        # now let's make sure (aka asset) the dir has no dataset
        …

        # now we download the datasets were properly saved  
        …

        # assert column names are in pt-br

        # translate and assert they are in en-us
        …
        
        # let's group the reimbursements (assert it, and that files by years are gone)
        …
```
</details>

luizcavalcanti:
Sure.

About the changes in data structures, only some column names had changed to camel-case (some didn't). Since we were performing case-sensitive filtering on those, I updated them. I didn't notice any difference in categories (except they corrected a extra blank space in one of them) or their ids. The biggest weirdness I faced was some lines in 2009 data that had 31 columns instead of 29 (a error_bad_lines=False was added on load_csv() to discard those lines).

Regarding tests, I completely agree it's a little wrong (:P) the way it is. I just didn't feel this was something to amend on this specific commit, but I'd say we should do it on this PR or in another one, soon enough.

jtemporal:
As requested WIP label removed ;)

serenata_toolbox/ceap_dataset.py
```diff
@@ -1,24 +1,23 @@
...
     def fetch(self):
-        urls = ['http://www.camara.gov.br/cotas/AnoAtual.zip',
-                'http://www.camara.gov.br/cotas/AnoAnterior.zip',
-                'http://www.camara.gov.br/cotas/AnosAnteriores.zip']
-        filenames = map(lambda url: url.split('/')[-1], urls)
+        base_url = "http://www.camara.leg.br/cotas/Ano-{}.csv.zip"
 
-        for url, filename in zip(urls, filenames):
-            zip_file_path = os.path.join(self.path, filename)
+        years = [n for n in range(2009, date.today().year+1)]
```
cuducos: We could have a constant `YEARS = [n for n in range(2009, date.today().year+1)]` in the beginning of the file, couldn't we? (relevant here and on line 37)

serenata_toolbox/ceap_dataset.py
```diff
@@ -30,15 +29,13 @@ def fetch(self):
 
 
     def convert_to_csv(self):
-        for filename in ['AnoAtual', 'AnoAnterior', 'AnosAnteriores']:
-            xml_path = os.path.join(self.path, '{}.xml'.format(filename))
-            csv_path = xml_path.replace('.xml', '.csv')
-            convert_xml_to_csv(xml_path, csv_path)
+        # deprecated but still here so we don't break poor Rosie (for now)
```
cuducos: ❤️

serenata_toolbox/reimbursements.py
```diff
@@ -41,7 +41,7 @@ def read_csv(self, name):
     @property
     def receipts(self):
         print('Merging all datasets…')
-        datasets = ('current-year.xz', 'last-year.xz', 'previous-years.xz')
+        datasets = ["reimbursements-{}.xz".format(n) for n in range(2009, date.today().year+1)]
```
cuducos: I'm not a big fan of long lines. If I was coding that probably I would extract it as a method decorated with `@property`:

```python
class Reimbursements:

    …
    @property
    @staticmethod
    def datasets():
        for year in range(2009, date.today().year+1):
            yield "reimbursements-{}.xz".format(year)
```

luizcavalcanti: I'm not really a python guy, so forgive me if pythonic stuff seems weird to me, I'm really new to all this. Can't we just create the YEARS constant, as suggested for CEAPDataset class and use it to reduce the line size?

cuducos: Yep, sure!

luizcavalcanti:
**Can you guys please put it again under "Work in Progress" label?** Congress changed the data structure, apparently. Tests are breaking again, several years have missing or changed categories (subquota description). I'll try to understand whats going on, if its temporary and try to make it more robust to those glitches/changes.

cuducos:
> Can you guys please put it again under "Work in Progress" label?

Done!

luizcavalcanti:
There was a change made on how subquota descriptions were translated. I don't know how pythonic it is, but it's fast enough, and most importantly, it does not break if something changes in the backend (descriptions changed and new subquotas were added in the past, and probably will again).

serenata_toolbox/ceap_dataset.py
```diff
@@ -50,103 +49,76 @@ def clean(self):
...
+        data.loc[data['subquota_number']==14,  ['subquota_description']] = 'Lodging, except for congressperson from Distrito Federal'
+        data.loc[data['subquota_number']==15,  ['subquota_description']] = 'Automotive vehicle renting or watercraft charter'
+        data.loc[data['subquota_number']==119, ['subquota_description']] = 'Aircraft renting or charter of aircraft'
+        data.loc[data['subquota_number']==120, ['subquota_description']] = 'Automotive vehicle renting or charter'
+        data.loc[data['subquota_number']==121, ['subquota_description']] = 'Watercraft renting or charter'
+        data.loc[data['subquota_number']==122, ['subquota_description']] = 'Taxi, toll and parking'
+        data.loc[data['subquota_number']==123, ['subquota_description']] = 'Terrestrial, maritime and fluvial tickets'
+        data.loc[data['subquota_number']==137, ['subquota_description']] = 'Participation in course, talk or similar event'
+        data.loc[data['subquota_number']==999, ['subquota_description']] = 'Flight ticket issue'
+
```
cuducos: I would refactor this part as:

```python
    subquotas = (
        (1, 'Maintenance of office supporting parliamentary activity'),
        (2, 'Locomotion, meal and lodging'),
        (3, 'Fuels and lubricants'),
        (4, 'Consultancy, research and technical work'),
        (5, 'Publicity of parliamentary activity'),
        (6, 'Purchase of office supplies'),
        (7, 'Software purchase or renting; Postal services; Subscriptions'),
        (8, 'Security service provided by specialized company'),
        (9, 'Flight tickets'),
        (10, 'Telecommunication'),
        (11, 'Postal services'),
        (12, 'Publication subscriptions'),
        (13, 'Congressperson meal'),
        (14, 'Lodging, except for congressperson from Distrito Federal'),
        (15, 'Automotive vehicle renting or watercraft charter'),
        (119, 'Aircraft renting or charter of aircraft'),
        (120, 'Automotive vehicle renting or charter'),
        (121, 'Watercraft renting or charter'),
        (122, 'Taxi, toll and parking'),
        (123, 'Terrestrial, maritime and fluvial tickets'),
        (137, 'Participation in course, talk or similar event'),
        (999, 'Flight ticket issue')
    )
    
    for code, name in subquotas:
        data.loc[data['subquota_number']==code, ['subquota_description']] = name
```

What do you think about it?

luizcavalcanti: I think it's much better. I'll do it that way.

luizcavalcanti:
I did no notebook on this, only compared with interactive python and some bash utils, if that count as analysis. In short, I found out that the data is the same between the API versions. 

About testing, what kind of tests do you foresee in this case? Is it worth to validate the translation itself? Or even make tests fail if there is new subquota categories?

cuducos:
> I did no notebook on this, only compared with interactive python and some bash utils, if that count as analysis. In short, I found out that the data is the same between the API versions.

I encourage you to do either share this analysis here (as a comment) or, even better: once this is merged, add a notebook to `serenata-de-amor` repo with the comparison between the two latest versions of `reimbursements.xz` (yours and current latest).

> About testing, what kind of tests do you foresee in this case? Is it worth to validate the translation itself? Or even make tests fail if there is new subquota categories?

I would use `unittest.mock` to avoid I/O basically. I don't need to actually run, for example, `urlretrieve` — I need to be sure it was called with the proper arguments. This means we don't depend on file system (mocking read/write methods) and network (mocking requests), i.e. faster tests, and tests restricted to what actually need to be tested.

These mocks are not trivial, so I don't push everyone to write them — unless anyone is up for pairing on that one of these days ; )

luizcavalcanti:
Looking into all that, @cuducos.

Just to make it public, found out yesterday that the congress is not sending the expenses issue date on the 2.0 dataset. PR on hold until they fix it. A complaint was filed to them, I'll post you guys and gals, once they do something about it.

luizcavalcanti:
Don't mind me, just rebasing with master to keep things saner when/if we ever merge this :P

cuducos:
Any news from them related to the missing `issue-date`?

luizcavalcanti:
The complaint is getting internal updates by the Camara staff, but it's invisible to me what those updates are. I can only hope they are talking about it and expect some public update soon.
screen shot 2017-05-08 at 18 36 52

Irio:
@luizcavalcanti Can you send me (privately is ok) the exact message you sent, with contact id? I can forward it directly to the team responsible for open data in the Chamber of Deputies.

luizcavalcanti:
Protocol: 170424-000119

"Gostaria de comunicar que a versão 2.0 dos arquivos de prestação de contas da cota para exercício de atividade parlamentar (CEAP), disponível em http://www2.camara.leg.br/transparencia/cota-para-exercicio-da-atividade-parlamentar/dados-abertos-cota-parlamentar, as datas de emissão (campo datEmissao) estão vindo vazias em todos os formatos (xml, csv, json e xlsx).

Estes dados estão disponíveis na versão 1.0 (AnoAtual.zip e AnosAnteriores.zip), mas não na versão 2.0. É importante que isso seja ajustado para que possamos usar o novo formato de acesso aos dados da CEAP, que é muito mais eficiente que o anterior."

luizcavalcanti:
A jupyter notebook was made to validate this PR. It has its own PR: https://github.com/datasciencebr/serenata-de-amor/pull/241

tests/test_chamber_of_deputies_dataset.py
```diff
@@ -11,37 +14,67 @@ class TestChamberOfDeputiesDataset(TestCase):
...
+        self.years = [n for n in range(2009, date.today().year + 1)]
+
 
     @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
             'Skipping integration test')
-    def test_fetch_saves_raw_files(self):
+    def test_fetch_translate_clean_integration(self):
+        # cleaning up previous tests
+        try:
+            os.remove(os.path.join(self.path, "datasets-format.html"))
```
cuducos: There is a `tearDown` method that runs after each test: wouldn't it be interesting to clean up there instead of in the beginning of a specific test? `tearDown` could just `shutil.rmtree(self.path)` for example.

jtemporal: Also if you want to clean up after you run all tests you can use `tearDownClass` decorated with `@classmethod` ;)

luizcavalcanti: I believe tearDown are inadequate since we need to guarantee deletion before we run the tests (in case the last execution was cancelled/aborted) even when `tearDown` was not executed last time. I guess `setUp` should be more appropriate, maybe with a test for `os.environ.get('RUN_INTEGRATION_TESTS') != '1'`, so we don't execute them in (probable) future unit tests.

luizcavalcanti: Oh, and self.path in MacOS points to a /var/folder/... subdirectory with a ton of other stuff besides serenata files. So I guess it's a no-no

cuducos: > Oh, and self.path in MacOS points to a /var/folder/... subdirectory with a ton of other stuff besides serenata files. So I guess running a rmtree() on it is a no-no

True story… let's use `tempfile.mksdtemp()` then ; )

> I believe `tearDown` are inadequate since we need to guarantee deletion before we run the tests (in case the last execution was cancelled/aborted) even when tearDown was not executed last time.

Makes sense, but I have never heard about this reflection before. Maybe extract the cleanup logic to an specific method and call it in `setUp` (just in case) and `tearDown`.

luizcavalcanti: While replacing gettempdir() for mkdtemp(), I found out it creates a new, different, "random" temp dir for every run, avoiding the need to clean up that upfront. Nice.

jtemporal:
just passing by to mention that is required a version bump here

cc @cuducos

cuducos:
> just passing by to mention that is required a version bump here

Thanks! I was so enthusiastic about it I was already forgetting… @luizcavalcanti in `setup.py` you can make an [micro version bump](https://github.com/datasciencebr/serenata-toolbox#development) (if #71 is merged soon with `10.0.2` yours could be `10.0.3` I guess).

Also if you could check merge conflicts it helps a lot ; ) Otherwise I'll sort that later ; )

luizcavalcanti:
Version bumped and whole branch rebased to current master

cuducos:
Yay… that's a great contribution, @luizcavalcanti! Many thanks

PR #44, opened 2017-04-26
Fix namespace

### LINKED ISSUE #43: Fix namespace of the package
There are 3 ways of setting up a python package as [Python Package User Guide](https://packaging.python.org) gracefully shows and I copy here:

1. Use native namespace packages. This type of namespace package is defined in PEP 420 and is available in Python 3.3 and later. This is recommended if packages in your namespace only ever need to support Python 3 and installation via pip.
1. Use pkgutil-style namespace packages. This is recommended for new packages that need to support Python 2 and 3 and installation via both pip and python setup.py install.
1. Use pkg_resources-style namespace packages. This method is recommended if you need compatibility with packages already using this method or if your package needs to be zip-safe.

The toolbox is structured according to the second name spacing type, which is as follows:
```
setup.py
mynamespace/
    __init__.py  # Namespace package __init__.py
    subpackage_a/
        __init__.py  # Sub-package __init__.py
        module.py
```
So for imports to work properly we need to fix the `serenata_toolbox/__init__.py` file to contain
```python
__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```
as it is explained [here](https://packaging.python.org/namespace_packages/#pkg-resources-style-namespace-packages).

jtemporal:
this PR should close #43

### DIFF
--- a/serenata_toolbox/__init__.py
+++ b/serenata_toolbox/__init__.py
@@ -1,3 +1,1 @@
-from .ceap_dataset import CEAPDataset
-from .datasets import fetch_latest_backup
-from .reimbursements import Reimbursements
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
--- a/setup.py
+++ b/setup.py
@@ -28,5 +28,5 @@ setup(
     name='serenata-toolbox',
     packages=['serenata_toolbox'],
     url=REPO_URL,
-    version='8.0.0'
+    version='8.0.1'
 )


cuducos:
I understand the theory underneath #43 but is there a more pragmatic use case exemplifying this issue? I ask that because **maybe** this fix changes the public API of the package — and that would not be a _minor_ version bump…

jtemporal:
Okay, so
> maybe this fix changes the public API of the package — and that would not be a minor version bump

It does not.

Before this alteration, if you run any setup that needs the toolbox you get the following error:
photo_2017-04-26_14-09-53
because there is a failure in the namespace so, basically Python was ignoring the `datasets` submodule.

with that correction everything will be back on track.

jtemporal:
thank you @rodolfo-viana for catching this today ;)

cuducos:
> It does not.

🎉

rodolfo-viana:
> thank you @rodolfo-viana for catching this today ;)

<3

PR #47, opened 2017-04-29
Add old config.ini message

cuducos:
Friendly ask users to replace, if needed, `s3-sa-east-1` (Serenata old default for _Region_ in `config.ini`) by `sa-east-1` (semantically correct: `s3` is not a `region`, is a service).


jtemporal:
hey @cuducos I would love to merge this PR, but some checks aren't passing, would you mind checking this out?

cuducos:
Shame on me. I was totally sleepy when I coded that. Anyway… fixed by a4a5d72

PR #51, opened 2017-05-03
Refactor toolbox structure to prepare for Federal Senate dataset

anaschwendler:
As we were doing with Rosie, we are now preparing the toolbox to receive new datasets.
What need to be done after that:
- [ ] Add the script to get the Federal Senate datasets.
- [ ] Add the datasets to Amazon S3


jtemporal:
I believe this is a `major` version bump: `9.0.0` since it changes the API @cuducos what do you think?

jtemporal:
would consider opening up issues with:
>- [ ] Add the script to get the Federal Senate datasets.
>- [ ] Add the datasets to Amazon S3

anaschwendler:
@jtemporal Yes, when it get merged, I'll be opening those issues.
Right after it gets merged I'll be adapting Rosie to new toolbox, and then get back to those issues

Issue #23, opened 2017-01-25
Improve documentation


jtemporal:
Now with [this issue](https://github.com/datasciencebr/serenata-de-amor/issues/143) being worked on we need a better documentation on how each toolbox submodule works.

igoragatti:
May I suggest activating the Wiki on Github so that documentation can be directly added to it?

PR #53, opened 2017-05-08
Introducing federal senate script

anaschwendler:
This is the first test and script for fetching the Federal Senate datasets.
Soon we will be able to add the new datasets to Amazon and use them normally.

I don't think it needs more cleaning, but I will be studying it later.
The translation and compression tasks are already working, an everything was developed using TDD.
Feel free to help :)


serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+import pandas as pd
+
+class FederalSenateDataset:
+    URL = 'http://www.senado.gov.br/transparencia/LAI/verba/{}.csv'
+
+    def __init__(self, path):
+        self.path = path
+
+    def fetch(self):
+        urls = [self.URL.format(year) for year in range(2008, 2018)]
```
cuducos: This hardcoded `2018` seems like a problem. We could use something like that to avoid bumping this year every new years eve:

```python
from datetime import date

…

    def fetch(self):
        next_year = date.today().year + 1
        urls = [self.URL.format(year) for year in range(2008, next_year)]
        …
```

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        filename_from_url = lambda url: 'federal-senate-{}'.format(url.split('/')[-1])
+        filenames = map(filename_from_url, urls)
+
+        for url, filename in zip(urls, filenames):
```
cuducos: <3

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+
+class FederalSenateDataset:
+    URL = 'http://www.senado.gov.br/transparencia/LAI/verba/{}.csv'
+
+    def __init__(self, path):
+        self.path = path
+
+    def fetch(self):
+        urls = [self.URL.format(year) for year in range(2008, 2018)]
+        filename_from_url = lambda url: 'federal-senate-{}'.format(url.split('/')[-1])
```
cuducos: Having a `lambda` this way is not a best practice, but I can handle that. The real problem here is that using [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) and [`os.path.basename`](https://docs.python.org/3/library/os.path.html#os.path.basename) would be way safer ; )

Something among these lines:

```python
from urllib.urlparse import urlsplit

…

    def fetch(self):
        …
        url_paths = (urlsplit(url).year for url in urls)
        filenames = map(os.path.basename, url_paths)
        …
```



os.path.basename

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        filenames = map(filename_from_url, urls)
+
+        for url, filename in zip(urls, filenames):
+            csv_file_path = os.path.join(self.path, filename)
+            urlretrieve(url, csv_file_path)
+
+    def translate(self):
+        filenames = ['federal-senate-{}.csv'.format(year) for year in range(2008, 2018)]
```
cuducos: Oops… this is the 2nd time I see this `range(2008, 20180)`! Let's make it a constant (at least a class constant like `URL`).

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        for filename in filenames:
+            csv_path = os.path.join(self.path, filename)
+            self.__translate_file(csv_path)
+
+    def __translate_file(self, csv_path):
+        output_file_path = csv_path \
```
cuducos: I'm afraid this line break here is unnecessary…

tests/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+        self.subject = FederalSenateDataset(self.path)
+
+    @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
+            'Skipping integration test')
+    def test_fetch_saves_raw_files(self):
+        self.subject.fetch()
+        names = ['federal-senate-{}.csv'.format(year) for year in range(2008, 2018)]
+        for name in names:
+            file_path = os.path.join(self.path, name)
+            assert(os.path.exists(file_path))
```
cuducos: What if files were there from another round of tests? This test suite need either a `tearDown` to clean up or to use mocks to avoid writing to file system IMHO…

tests/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
+            'Skipping integration test')
+    def test_translate_creates_english_versions_for_every_csv(self):
```
cuducos: We test for files existence/absence, but not for translations themselves — this might not be a priority right now, but at least we should be aware of that and register that as an issue.

tests/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+        self.subject.translate()
+        names = ['federal-senate-{}.xz'.format(year) for year in range(2008, 2018)]
+        for name in names:
+            file_path = os.path.join(self.path, name)
+            assert(os.path.exists(file_path))
+
+if __name__ == '__main__':
+    main()
```
cuducos: We don't need that I guess — we use a [test finder](https://github.com/datasciencebr/rosie/blob/master/rosie.py#L35) ; )

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: done

tests/test_federal_senate_dataset.py
anaschwendler: done

tests/test_federal_senate_dataset.py
anaschwendler: I was thinking about it yesterday, we don't test if the translation was successfully made. I was thinking about opening an issue with that, and let it to be done after the migration.

We need to do it in `chamber_of_deputies` module too.

tests/test_federal_senate_dataset.py
anaschwendler: I completely agree that we need to add a `tearDown` to clean up, but I don't know how to do it, examples?

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: Ok, I made a constant for first and next year :)
Adapted to what you suggested, soon will be pushing it.

tests/test_federal_senate_dataset.py
anaschwendler: We already need this to run the tests individually. 
I will keep that part :)

tests/test_federal_senate_dataset.py
cuducos: Let's say your test creates a bizarre file: `/tmp/my-bizarre-and-beloved-test-side-effect.pkl`:

```python
class TestSomething(TestCase):

    def setUp(self):
        self.path = gettempdir()
        self.file_path = os.path.join(self.path, 'my-bizarre-and-beloved-test-side-effect.pkl')
        …

    def tearDowm(self):
        os.remove(self.file_path)

    def test_something(self):
        pass  # something that creates the bizarre file
```

Everything on `setUp` runs before every test. Everything in `tearDown` runs after every test. So you can be sure that after every test the bizarre file is deleted. Surely it might need an is `os.path.exist`, or a `try`/`except` if the file wasn't created in every test… but this is the general idea.

Makes sense?

anaschwendler: It makes sense, the only thing I need to know is: I need to do it for all files I created in the test, or this only line destroy all of them?

cuducos: In the example [`os.remove`](https://docs.python.org/3.6/library/os.html#os.remove) will only remove the file from the path passed as an argument ; ) You have to pass the files you need to clean up ; )

anaschwendler: thank you <3

cuducos:
Ok, so now there's two things pending, right?

- [ ] Resolve conflicts
- [ ] Merge datasets by year in a single dataset

anaschwendler:
Yes, that is what I will be doing now :)

anaschwendler:
On hold until we get the [full analysis of the datasets](https://github.com/datasciencebr/serenata-de-amor/pull/231), to clean it in the right way.

So we get back, to go further ¯\_(ツ)_/¯

anaschwendler:
Unholding this, because finally we decided something.
All the study (is here)[https://github.com/datasciencebr/serenata-de-amor/pull/231].

We decided that we will only clean the `date` and the `cnpj_cpf` fields and make another exploratory works after get the basic done.

Thanks @jtemporal and @cuducos for all feedback, everything is close to an end.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+        self.path = path
+
+    def fetch(self):
+        for year in range(self.FIRST_YEAR, self.NEXT_YEAR):
+            url = self.URL.format(year)
+            file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
+            urlretrieve(url, file_path)
+
+    def translate(self):
+        filenames = ['federal-senate-{}.csv'.format(year) for year in range(self.FIRST_YEAR, self.NEXT_YEAR)]
```
cuducos: I'd add something as `YEAR_RANGE` as a class constant to avoid repeating this `range(…)`.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+            'Passagens aéreas, aquáticas e terrestres nacionais':
+                'National air, water and land transport',
+            'Serviços de Segurança Privada':
+                'Private Security Services'
+        }
+
+        categories = [categories[cat]
+                      for cat in data['expense_type'].cat.categories]
+        data['expense_type'].cat.rename_categories(categories,
+                                                           inplace=True)
```
jtemporal: 🎉 would remove those extra spaces before `inplace=True` to vertically align to `categories`

serenata_toolbox/federal_senate/federal_senate_dataset.py
jtemporal: I like this idea :)

anaschwendler: The step 2 is done here! <3
We merged the datasets already and cleaned up :)

anaschwendler: I'll do this `YEAR_RANGE` by now! 
Thanks for that!

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: Done! 🎉

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: `YEAR_RANGE` done too, great idea, thank you! <3

anaschwendler:
Checked all @cuducos and @jtemporal suggestions.
Checking [step 2 of 2](https://github.com/datasciencebr/serenata-toolbox/pull/53#issuecomment-300828176) because its done with `clean()` method.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+    YEAR_RANGE = range(self.FIRST_YEAR, self.NEXT_YEAR)
+
+    def __init__(self, path):
+        self.path = path
+
+    def fetch(self):
+        for year in YEAR_RANGE:
+            url = self.URL.format(year)
+            file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
+            urlretrieve(url, file_path)
```
lipemorais: I would like to be able feel protect by tests even if I'm offline, so mock urlretrieve will help with it and let us improve the tests in some others aspects.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+            dataset = pd.concat([dataset, data])
+
+        dataset['date'] = pd.to_datetime(dataset['date'], errors='coerce')
+        dataset['cnpj_cpf'] = dataset['cnpj_cpf'].str.replace(r'\D', '')
+
+        dataset.to_csv(reimbursement_path, compression='xz', index=False, encoding='utf-8')
+
+        return reimbursement_path
+
+    def __translate_file(self, csv_path):
```
lipemorais: This internal method looks a little big, could we break it to make the parts smaller and more meaningful and clear?

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+                'Publicity of parliamentary activity',
+            'Locomoção, hospedagem, alimentação, combustíveis e lubrificantes':
+                'Locomotion, lodging, food, fuels and lubricants',
+            'Passagens aéreas, aquáticas e terrestres nacionais':
+                'National air, water and land transport',
+            'Serviços de Segurança Privada':
+                'Private Security Services'
+        }
+
+        categories = [categories[cat] for cat in data['expense_type'].cat.categories]
```
lipemorais: At this line looks that is a translation from portuguese to english but it's changing state of categories variable here. It is making this code more complex than it could be. :/

May it could be done in a function just to give it a name and help with clarity.
 Show all reviewers

lipemorais:
I missing some unit tests. :( 

In order to have a smaller feedback cycle, I think that unit tests would be very helpful. Like test for each step of the script, for now I can think in tests for fetch using mocks for urlretrieve, clean and translate methods.

cuducos:
> I missing some unit tests. :( 

Me too, but [we've decided](https://github.com/datasciencebr/serenata-toolbox/pull/53#pullrequestreview-36862019) to let it go for now and we'll refactor tests later ; ) Wanna pair on that?

anaschwendler:
Hi @lipemorais!

There is a issue for unit tests, we are running to finish the script by now, but we are thinking about it.
The issue is here: https://github.com/datasciencebr/serenata-toolbox/issues/58

lipemorais:
> ... Wanna pair on that?

Hell yeah! <3

cuducos:
Anytime next week — I'll drop you a line ; )

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: So that is a thing that I was thinking about, but is a translation, so we need it :T

serenata_toolbox/federal_senate/federal_senate_dataset.py
anaschwendler: When it got merged, can you suggest something?
It will be pretty awesome to us! <3

lipemorais: Of course I can! I will be very happy to contribute for this awesome project. :)

anaschwendler:
The `clean()` method calls the `fetch()` and `translate()` the `.xz`file is create at the translation of the dataset.
Did you pulled the whole PR? 
I updated that yesterday :T

cuducos:
> The `clean()` method calls the `fetch()` and `translate()`, the `.xz`file is create at the translation of the dataset.

OMG we must pay attention to #23 so bad…

> Did you pulled the whole PR? 

Sure think. 

> So you need to run:
> `d.fetch()`
> `d.translate()`
> `d.clean() `

As you can see in my output I skip the `d.translate()`… testing it again and updating this thread.

anaschwendler:
> OMG we must pay attention to #23 so bad…
I know, that is a serious thing, we should work on it ASAP.

> As you can see in my output I skip the `d.translate()`… testing it again and updating this thread.
Thank you :)

anaschwendler:
I just noticed that what I said was kinda wrong.
Fixing it:
@cuducos 
The `fetch`, `translate` and `clean` methods are uniques, and they do specific things.
`fetch` get the datasets
`translate` translate the columns and the categories
`clean` concatenate the datasets and clear the `date` and `cnpj_cpf` fields.

So, what I was trying to say was that in my test, to test the `clean` method, I run `fetch()` and `translate()` after running `clean()` 

Does it makes sense right now?

lipemorais:
It make sense for me, but looks that all them could be testeds in clean as a journey. It helps also with don't repeat yourself principle and we will have less time ruining tests but the same coverage.

cuducos:
Ok. This is *so* confusing.

* The `clean` method merges the thing?
* The `translate` method converts the thing to `.xz`?

We really refactor to

1. Make methods more atomic (UNNIX philosophy: do one thing and one thing well) — this will lead us to five methods instead of three: something like `fetch`, `clean`, `translate`, `convert_to_xz` and `merge`
2. End up with meaningful and less confusing method names
3. Reflect this changes in `chamber_of_deputies`

If you agree would you mind opening an issue about it?

Besides that now that I ran thing i the right order (`fecth`, `translate` and `clean`) it works merging it 🎉

anaschwendler:
> The clean method merges the thing?

Yes, it does merge all datasets, with an `.xz` compression

> The translate method converts the thing to .xz?

Yes, it translate each one of the datasets fetched from senate website.

> If you agree would you mind opening an issue about it?

No, I totally agree with all of that, we really need to refactor it in `chamber_of_deputies` too, but I don't know what is the best way.

@lipemorais  has raised a flag to help, just can't wait to see the results :)

Thanks for merging it <3

PR #56, opened 2017-05-11
add zip_safe=False to setup.py

luzfcb:
The zip_safe flag can be used to force or prevent zip Archive creation. In general you probably don’t want your packages to be installed as zip files because some tools do not support them and they make debugging a lot harder.

To maintain sanity, this flag should be included in all the setup.py files of all datasciencebr python packages .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ setup(
         'Topic :: Utilities',
     ],
     description='Toolbox for Serenata de Amor project',
+    zip_safe=False,
     install_requires=[
         'aiofiles',
         'aiohttp',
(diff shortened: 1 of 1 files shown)


cuducos:
Great! Would mind bumping version before merging? Probably a minor change ; )

cuducos:
Many thanks ; )

luzfcb:
It may also be necessary to include `include_package_data=True`.  `include_package_data` tells setuptools to look for a `MANIFEST.in` file and install all the entries that match as package data.

django has a good [MANIFEST.in](https://github.com/django/django/blob/master/MANIFEST.in) example

cuducos:
Is this an issue for another PR or are you planning to bringing it in in this one? If that's the case I'll label it as WIP (work in progress) and you can work on that with no hurry ; )

luzfcb:
@cuducos `include_package_data=True` is a problem to another pull-request.

I found `zip_safe=False` error when try to solve serenata-toolbox incompatibility with `dependency_links` option of setup.py when I was making questions to setuptools / pip developers, to solve some things in https://github.com/datasciencebr/rosie/pull/44

PR #62, opened 2017-05-16
Fix bump version to federal senate script

anaschwendler:


cuducos:
As it added stuff to the public API I think it's a minor (not micro) version bump — i.e. `9.1.0`, nor `9.0.4` ; )

anaschwendler:
Aw thanks, forgot about that!

PR #68, opened 2017-05-18
Refactoring federal senate dataset

lipemorais:
This is in progress, I just created it earlier to have earlier feedbacks.

The purpose of this PR is make some improvements on tests and code of federeal_senate_dataset.

I created a new kind of test called **journey test** to test the real workflow even taking more time. Would like to receive some feedbacks to check if it's good or not for the project.
At the first time that I executed it, took more than 8 minutes now it takes almost 5:30 minutes. Se output below:
```
(.toolbox-env) ➜  fork-toolbox git:(master) ✗ python -m unittest tests/journey/test_federal_senate_dataset.py
.
----------------------------------------------------------------------
Ran 1 test in 338.959s

OK
(.toolbox-env) ➜  fork-toolbox git:(master) ✗
````

I also added a mocked unit test to improve the feedback cycle.

To run the journey test you can use the code below:
_Do not forget to export env var RUN_INTEGRATION_TESTS with value 1 to do not have this test skipped_
```
python -m unittest tests/journey/test_federal_senate_dataset.py
```

To run the unit test use this:
```
python -m unittest tests/test_federal_senate_dataset.py
```

@anaschwendler is the person with more knowledge about this code, her feedbacks will be very helpful. Other people that can help are @Irio @jtemporal @cuducos. If you are not one of them feel free to provide me some feedbacks.

**TODO**

- Add unit test for translate
- Add unit test for clean
- Improve method names

**Nice To Have**

- Refactor `fetch, translate, clean` **to** `fetch, translate, clean, convert, merge`
- Reflect this changes in chamber_of_deputies

### CODE REVIEW

**Comment by jtemporal** [comment]
>This is in progress

WIP label added, we can take it out whenever just ping us here =)

**Comment by anaschwendler** [comment]
Hi @lipemorais !
Thank you very much for this refactor, I'll be looking for that closely :)

In my opinion the only thing we have to keep, is the way that things works and some structures to keep everything working on the other places: `rosie` and `serenata-de-amor`.

So we need to keep the main methods: `fetch()`, `translate()` and `clean()` with its features. Everything besides that can be changed!

:tada:

**Comment by cuducos** [comment]
> So we need to keep the main methods: `fetch()`, `translate()` and `clean()` with its features. Everything besides that can be changed!

As [I've said elsewhere](https://github.com/datasciencebr/serenata-toolbox/pull/53#issuecomment-301807098) this API is terrible. `translate` is translating labels **and** changing the file extensions, `clean` is cleaning data **and** merging things. I definitively think this is a terrible API. It's confusing, it's not semantically meaningful, it's hard to test and hard to maintain.

I think TDD can help us to restructure this API, working with stream of data, reducing temporary intermediary files (like `federal_senate-2008.xz`), so we fetch original files, do all treatment needed and finally write the results. If memory is an issue this an awesome opportunity to [bring Dask in](https://github.com/datasciencebr/rosie/issues/24#issuecomment-279711878) — and get another leverage: to risk get rid of Rosie's memory issues.

@anaschwendler can you list exactly what we're gonna break if we change the input and output of these methods?

**Comment by anaschwendler** [comment]
So, @cuducos 
Hm, it is ok to refactor all the API, but we may need some time to reflact it in the other repositories, I can help with that and we need to do it very carefully.

> @anaschwendler can you list exactly what we're gonna break if we change the input and output of these methods?

For the Federal Senate we can work on it, because it is not on production, but for `chamber_of_deputies` it changes all its structure.

> and get another leverage: to risk get rid of Rosie's memory issues.

For me it looks perfect, and we can work together on it.

**Comment by cuducos** [comment]
Yay! Next week I can start working on that. If anyone is willing to study how embedded are those methods in the other repos architecture, that would be awesome. Specially `translate` and `clean` — `fetch` is good so far IMHO.

File: tests/journey/test_federal_senate_dataset.py [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TestJourneyFederalSenateDataset(TestCase):
+    def setUp(self):
+        self.path = gettempdir()
+        self.subject = FederalSenateDataset(self.path)
+
+    @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
+            'Skipping integration test')
+    def test_journey_federal_senate_dataset(self):
```
lipemorais: Yes. This journey test will do the whole workflow.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,79 @@
...
+        return reimbursement_path
 
-        filenames = ['federal-senate-{}.xz'.format(year) for year in self.YEAR_RANGE]
+    def cleanup_dataset(self, dataset):
+        dataset['date'] = pd.to_datetime(dataset['date'], errors='coerce')
+        dataset['cnpj_cpf'] = dataset['cnpj_cpf'].str.replace(r'\D', '')
+
+        return dataset
+
+    def merge_files(self, filenames):
```
lipemorais: @cuducos ;)

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,79 @@
...
         reimbursement_path = os.path.join(self.path, 'federal-senate-reimbursements.xz')
+        cleaned_merged_dataset.to_csv(reimbursement_path,
+                                      compression='xz',
+                                      index=False,
+                                      encoding='utf-8')
+
+        return reimbursement_path
 
-        filenames = ['federal-senate-{}.xz'.format(year) for year in self.YEAR_RANGE]
+    def cleanup_dataset(self, dataset):
```
lipemorais: @cuducos ;)

jtemporal:
@lipemorais In light of the last commits I wonder if wouldn't be wiser to generate the `.xz` from the `.csv` instead of having to commit them as fixtures?

lipemorais:
@jtemporal it would be good but make the coupling between the tests biggers, something like clean will just work after translate has runned. This would let the tests more slow and not so unit as it should be.

In order to have less fixtures I'm trying to make the tests uses less fixtures as possible, I will commit it soon.

lipemorais:
@anaschwendler @jtemporal @cuducos 
I need some help to fix to understand why is this test failing in Travis CI. :(

This is my last test for this PR.

The next step are more **nice to haves**

cuducos:
```
No such file or directory: 'tests/fixtures/csv/federal-senate-2008.xz'
```

I think it fails because when you instantiate something like `FederalSenateDataset('tests/fixtures/csv/', …)` you won't be able to access any `.xz` (because they're in `tests/fixtures/xz/`).

lipemorais:
@cuducos Actually when you call translate it generates the file with .xz extension. If you run it locally you will see it.

cuducos:
> @cuducos Actually when you call translate it generates the file with .xz extension. If you run it locally you will see it.

I know. But does it generates the `.xz` in which folder: `tests/fixtures/csv/` or `tests/fixtures/xz/`? That's my point.

lipemorais:
It generates `.xz` file in csv folder: `tests/fixtures/csv/ `.

cuducos:
> It generates `.xz` file in csv folder: `tests/fixtures/csv/ `.

Hum… confusing, but I got it. I have the impression we could abuse of mocks to avoid that mess in terms of file system. Unit tests to test small atomic stuff, and one integration test that would do everything form scratch…

lipemorais:
> Hum… confusing, but I got it.

Yeah, it's confusing. For the first step of this refactoring I trying to isolate network tests, a next step would be do the same with file handling tests.

I'm worried about `works in my machine problem`, I do not find a simple way to reproduce what is happening in Travis CI. It will be necessary soon or later, would be good be able to reproduce it.

I do like the idea of use mocks, it  will make the test better. Could you a little more about it?

cuducos:
>  Could you a little more about it?

Some examples:

* `fetch`: we don't need to actually run `urlretrieve`, we assume `urlretrieve` just works (Guido must have tested it hahaha) and must assert it was called x times with the proper arguments — so we mocked it and check for these calls, not for files written in the file system
* `translate`: we test `__translate_file` with a single unit test and them in the `translate` method what is relevant is to know if `self.__translate_file` was called x times with the proper argument
* `__translate_file`: we mock `read_csv` setting a pandas DataFrame we build in tests (in memory, not as a file) as a `return_value` and check if its keys/headers are actually translated
* Etc…

This way we don't need fixtures at all for unit tests. And for integration tests we actually download data and check if the journey succeeds from the scratch.

Does that make sense?

jtemporal:
Also minor comment on paths: Try to use the `os.path.join` to build the paths strings and avoid breaking things on windows ;)

lipemorais:
> Does that make sense?
@cuducos I do agree with everything that you just said.

> fetch: we don't need to actually run urlretrieve, we assume urlretrieve just works (Guido must have tested it hahaha) and must assert it was called x times with the proper arguments — so we mocked it and check for these calls, not for files written in the file system

Actually this is done in [tests/test_federal_senate_dataset.py:18](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L18)

> translate: we test __translate_file with a single unit test and them in the translate method what is relevant is to know if self.__translate_file was called x times with the proper argument

This is a great idea but now that translate test is working I think that this can wait a little.

> __translate_file: we mock read_csv setting a pandas DataFrame we build in tests (in memory, not as a file) as a return_value and check if its keys/headers are actually translated

Agree with it. We can create a issue just refactor translate. What do you think?

> This way we don't need fixtures at all for unit tests. And for integration tests we actually download data and check if the journey succeeds from the scratch.

With less fixtures we will have tests running even faster but since at the beginning was necessary 8:30 minutes and now we can run it in 6:32 seconds. I would like to merge this PR and let more improvements for a next one.

@jtemporal Thank you for your feedback. Where could I apply `os.path.join` in this PR to keep this code Windows friendly?

To finish the mystery. Travis was right, I made a mistake during a refactor when I changed a variable name. Here is the [commit](https://github.com/lipemorais/serenata-toolbox/commit/63292b8188cf318eb603a1b2b3d8cf09502d3515) that makes it work again.

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,37 @@
...
+        # fetch_saves_raw_files
+        self.subject.fetch()
+        federal_senate_csv_files = ['federal-senate-{}.csv'.format(year) for year in range(self.subject.FIRST_YEAR, self.subject.NEXT_YEAR)]
```
jtemporal: I would create an variable that has the year range just so you won't have to repeat `range(self.subject.FIRST_YEAR, self.subject.NEXT_YEAR)` and on line 26

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,37 @@
...
+        for federal_senate_csv_file in federal_senate_csv_files:
+            file_path = os.path.join(self.path, federal_senate_csv_file)
+            self.assertTrue(os.path.exists(file_path), "fetch_saves_raw_files")
```
jtemporal: Also noted inconsistencies on the use of double quotes and single ones... It would be nice to stick to one of them 💅

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,93 @@
...
 
+
 class TestFederalSenateDataset(TestCase):
-    def setUp(self):
+    @classmethod
+    def setUpClass(cls):
+        cls.expected_files = ['federal-senate-2008.csv',
+                              'federal-senate-2009.csv']
+
+    @patch("serenata_toolbox.federal_senate.federal_senate_dataset.urlretrieve")
```
jtemporal: again mix of double and single quote marks 💅

jtemporal:
If you run tests on windows the ones that have hardcoded path will fail =)

>Where could I apply os.path.join in this PR to keep this code Windows friendly?

so, wherever there is a path ;)
- [ ] [L43](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L43)
- [ ] [L55](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L55)
- [ ] [L74](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L74)
- [ ] [L86](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L86)
- [ ] [L92](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/test_federal_senate_dataset.py#L92)

cuducos:
`os.path.join` FTW ; )

lipemorais:
@jtemporal Thank you for feedbacks, I made the changes that you suggested and would like to see if now it's ok for you.

gustavo-freitas:
@lipemorais @cuducos @jtemporal hello everybody! I'm totally new here! @lipemorais can you point  me out exactly where is the journey test that takes around 6/7 minutes?

lipemorais:
Hey, @gustavo-freitas ! I'm glad to have you here.

This is the [journey test](https://github.com/lipemorais/serenata-toolbox/blob/master/tests/journey/test_federal_senate_dataset.py)

To make it run use:
Obs.: Python3 is expected and may some dependencies needed to run this test. 
```
python -m unittest tests/journey/test_federal_senate_dataset.py
```

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,92 @@
-import os
 from tempfile import gettempdir
-from unittest import main, skipIf, TestCase, TestLoader
+from unittest import TestCase
+from unittest.mock import patch
+import pandas as pd
+import os
```
jtemporal: It is very important to pay attention to the imports order and separating lines between types of imports, this imports should be:
```
import os
from tempfile import gettempdir
from unittest import TestCase
from unittest.mock import patch

import pandas as pd
```
Refer to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) for further instructions.

Particularly right now I'm using [isort](https://pypi.python.org/pypi/isort) to organize my imports, I believe @cuducos has other software he could recommend (can't remember the name)

tests/test_reimbursements.py
```diff
@@ -1,4 +1,4 @@
-from unittest import TestCase, main
+from unittest import TestCase
```
jtemporal: Also on the matter of imports, would you mind fixing these too?

``` 
from unittest import TestCase

import numpy as np
import pandas as pd
from numpy.testing import assert_array_equal

 from serenata_toolbox.chamber_of_deputies.reimbursements import Reimbursements
```

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,37 @@
+import os
+from tempfile import gettempdir
+from unittest import main, skipIf, TestCase
+
+
```
jtemporal: Also on the matter of imports: `from unittest import TestCase, main, skipIf` and there is an extra line here that should bem removed

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,37 @@
...
+        for federal_senate_xz_file in federal_senate_xz_files:
+            file_path = os.path.join(self.path, federal_senate_xz_file)
+            self.assertTrue(os.path.exists(file_path), 'translate_creates_english_versions_for_every_csv')
+
+        # clean_creates_a_reimbursements_file
+        self.subject.clean()
+        file_path = os.path.join(self.path, 'federal-senate-reimbursements.xz')
+        self.assertTrue(os.path.exists(file_path), 'clean_creates_a_reimbursements_file')
+
+if __name__ == '__main__':
```
jtemporal: there should be two lines before this if statement =) (for this one I used [flake8](http://flake8.pycqa.org/en/latest/))

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,92 @@
...
+
+        translated_files, not_found_files = self.subject.translate()
+
+        for translated_file, expected_file in zip(
+                translated_files, expected_files):
+
+            self.assertIn(expected_file, translated_file)
+
+    def test_if_translation_happened_as_expected(self):
+        self.subject = FederalSenateDataset(os.path.join('tests/fixtures/csv/'), 2008, 2009)
```
jtemporal: Here the idea of using `os.path.join` is so that you do not have to set the path with the `/` separator like: `os.path.join('tests', 'fixtures', 'csv')`.

Why is that? Because the separator used in file paths on Windows is `\` and not `/` as Unix based systems ;)

So the output of `os.path.join('tests', 'fixtures', 'csv')` on:
- Windows would be `tests\fixtures\csv\`
- Unix based would be `tests/fixtures/csv/`

And this is only one of the reasons to use `os.path.join` =)

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,92 @@
...
-    main()
+        translated_file_path = self.subject.path + 'federal-senate-2008.xz'
+        translated_federal_senate_2008 = pd.read_csv(translated_file_path,
+                                                     encoding='utf-8')
+
+        self.assertIsNotNone(translated_federal_senate_2008['year'],
+                             'expects \'year\' as column in this dataset')
+
+    def test_dataset_translation_failing_to_find_file(self):
+        self.subject = FederalSenateDataset(os.path.join('tests/fixtures/csv/'), 2007, 2008)
```
jtemporal: `os.path.join('tests', 'fixtures', 'csv')` again =)

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,92 @@
...
+
+        translated_files, not_found_files = self.subject.translate()
+
+        for not_found_files, expected_file in zip(
+                not_found_files, expected_files):
+
+            self.assertIn(expected_file, not_found_files)
+
+    def test_dataset_cleanup(self):
+        self.subject = FederalSenateDataset(os.path.join('tests/fixtures/xz/'), 2009, 2010)
```
jtemporal: `os.path.join('tests', 'fixtures', 'xz')`

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,92 @@
...
+
+        reimbursement_path = self.subject.clean()
+
+        self.assertEqual(
+            reimbursement_path,
+            os.path.join('tests/fixtures/xz/federal-senate-reimbursements.xz')
```
jtemporal: `os.path.join('tests', 'fixtures', 'xz', 'federal-senate-reimbursements.xz')`

tests/test_federal_senate_dataset.py
lipemorais: [Isort](https://pypi.python.org/pypi/isort) is awesome. I added a plugin on my Atom.

tests/journey/test_federal_senate_dataset.py
lipemorais: Nice!

lipemorais:
@jtemporal I made the fixes that you suggested and opened a [issue](https://github.com/datasciencebr/serenata-toolbox/issues/77) to make it work for the whole project too. Thank you for your feedbacks.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
+        retrieved_files = []
+        not_found_files = []
+
+        for year in self.year_range:
             url = self.URL.format(year)
             file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
-            urlretrieve(url, file_path)
+            try:
+                urlretrieve(url, file_path)
+                retrieved_files.append(file_path)
```
cuducos: I would move this line to the `else` block:

```python
try:
    urlretrieve(url, file_path)  # only code that might raise the exception
except Exception as exception:
    print('While fetching Seranata Toolbox not found file: {0} \n{1}'.format(file_path, exception))
    not_found_files.append(file_path)
else:
    retrieved_files.append(file_path)  # code to be executed if no exception was raised
```

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
+            except Exception as exception:
+                print('While fetching Seranata Toolbox not found file: {0} \n{1}'.format(file_path, exception))
```
cuducos: No need to number the `{…}` in the Python version we use (actually in anythig higher than Python 3.0. Can you cleanup?

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
+        filenames = self._filenames_generator('csv')
+        not_found_files = []
+        translated_files = []
+
         for filename in filenames:
             csv_path = os.path.join(self.path, filename)
-            self.__translate_file(csv_path)
+            try:
+                self._translate_file(csv_path)
+                translated_files.append(csv_path)
```
cuducos: Same as before… let's move it to an `else` block ; )

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
+            except Exception as exception:
+                print('While translating Seranata Toolbox not found file: {0} \n{1}'.format(csv_path, exception))
```
cuducos: Same as before: no numbering needed.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
 
-        filenames = ['federal-senate-{}.xz'.format(year) for year in self.YEAR_RANGE]
-        dataset = pd.DataFrame()
+        return reimbursement_path
 
-        for filename in filenames:
-            file_path = os.path.join(self.path, filename)
-            data = pd.read_csv(file_path, encoding = "utf-8")
-            dataset = pd.concat([dataset, data])
+    def _filenames_generator(self, extension):
```
cuducos: Grammar: `filename_generator` (no plural)

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,93 @@
...
+        return ['federal-senate-{0}.{1}'.format(year, extension) for year in self.year_range]
```
cuducos: Same as before ; )

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,34 @@
+import os
+from tempfile import gettempdir
+from unittest import TestCase, skipIf
+
+from serenata_toolbox.federal_senate.federal_senate_dataset \
+    import FederalSenateDataset
```
cuducos: [I don't care about PEP8 when it ends up making code harder to read](https://youtu.be/wf-BqAjZb8M). Let's go beyond 80 cols here?

tests/journey/test_federal_senate_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestJourneyFederalSenateDataset(TestCase):
+    def setUp(self):
+        self.path = gettempdir()
+        self.subject = FederalSenateDataset(self.path)
+
+    @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
```
cuducos: Is sounds counter intuitive: if `RUN_INTEGRATION_TESTS` is not set, `0`, `False`, `''` (empty string) then we actually **run** the integration tests?

lipemorais: No, if anything different of 1 is set this integration test don't run.

cuducos: Ow… true story, my bad!

cuducos:
Ok… for me just one last command: we don't have `fixtures` in the `test/fixtures` directory. We actually have full datasets there. Due to the strategy adopted in this PR I understand we need these files there. But this is bizarre:

```
$ du -h -d 1
 15M	./.git
 44K	./docs
280K	./htmlcov
256K	./serenata_toolbox
 20K	./serenata_toolbox.egg-info
 51M	./tests
 67M	.
```
What I mean is: the whole thing excluding `.git/` has 52Mb. 98% of the repo is in the `tests/fixtures/` directory:

```
$ du -h -d 1 tests/fixtures
 47M	tests/fixtures/csv
3.9M	tests/fixtures/xz
 51M	tests/fixtures
```

IMHO we can be more concise when it comes to fixtures. We can:

1. edit fixtures to have 2 lines each, for example (or more _only_ if we have a plausible reason to add more)
2. edit test routines to need only data from 2 years (e.g. 2008 and 2009)

This seams enough for testing purposes… what do you think @jtemporal?

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,102 @@
...
+
 class TestFederalSenateDataset(TestCase):
-    def setUp(self):
+    @classmethod
+    def setUpClass(cls):
+        cls.expected_files = ['federal-senate-2008.csv',
+                              'federal-senate-2009.csv']
+
+    @patch('serenata_toolbox.federal_senate.federal_senate_dataset.urlretrieve')
+    def test_fetch_files_from_S3(self, mockedUrlRetrieve):
```
cuducos: `mocked_url_retrieve` (camel case only for classes)

cuducos:
Sorry… that wasn't my last comment. The test suit is leaving a residual file behind, we need to fix that before merging:

```console
~/serenata-toolbox(lipemorais-master) » git status                                                                       
On branch lipemorais-master
nothing to commit, working tree clean

~/serenata-toolbox(lipemorais-master) » python -m unittest discover tests                                                
s..........................Could not find config.ini file.
You need Amazon section in it to interact with S3
(Check config.ini.example if you need a reference.)
.You need an Amazon section in config.ini to interact with S3 (Check config.ini.example if you need a reference.)
.Could not find config.ini file.
You need Amazon section in it to interact with S3
(Check config.ini.example if you need a reference.)
.Could not find config.ini file.
You need Amazon section in it to interact with S3
(Check config.ini.example if you need a reference.)
.....It looks like you have an old version of the config.ini file. We do not need anymore the service (s3) appended to the region (sa-east-1). Please update your config.ini replacing regions like `s3-sa-east-1` by `sa-east-1`.
.....While translating Seranata Toolbox not found file: tests/fixtures/csv/federal-senate-2007.csv
File b'tests/fixtures/csv/federal-senate-2007.csv' does not exist
..While fetching Seranata Toolbox not found file: /var/folders/lr/gd0yq9m90rv_gz682gqt7ylc0000gn/T/federal-senate-2007.csv
HTTP Error 404: Not Found
..Loading receipts.csv…
Dropping rows without document_value or reimbursement_number…
Grouping dataset by applicant_id, document_id and year…
Gathering all reimbursement numbers together…
Summing all net values together…
Summing all reimbursement values together…
Generating the new dataset…
.
----------------------------------------------------------------------
Ran 45 tests in 5.566s

OK (skipped=1)

~/serenata-toolbox(lipemorais-master*) » git status                                                                      
On branch lipemorais-master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	tests/fixtures/csv/federal-senate-2008.xz

nothing added to commit but untracked files present (use "git add" to track)

~/serenata-toolbox(lipemorais-master*) »
```

tests/test_federal_senate_dataset.py
```diff
@@ -1,41 +1,102 @@
...
-            assert(os.path.exists(file_path))
-
-    @skipIf(os.environ.get('RUN_INTEGRATION_TESTS') != '1',
-            'Skipping integration test')
-    def test_clean_creates_a_reimbursements_file(self):
-        self.subject.fetch()
+        retrieved_files, not_found_files = self.subject.fetch()
+
+        self.assertTrue(mockedUrlRetrieve.called)
+        self.assertEqual(mockedUrlRetrieve.call_count, 10)
```
cuducos: Maybe next year this test will fail as we're not providing a controlled `LAST_YEAR`…

tests/test_federal_senate_dataset.py
lipemorais: Hahaha, it's a lot o Java in my mind.

cuducos: Addressed that in 598545a0f1c479503fc6c6bd99c6c34cdd496801, feel free to bring that to your branch ; )

tests/test_federal_senate_dataset.py
cuducos: Addressed that in b1e094c2da4c7c0dc2e8cc7b19346cf2ca677a95 (feel free to bring that to your branch)

cuducos:
In sum:

* I addressed the fixtures size (and some minor details) in a [another branch based on this PR](https://github.com/datasciencebr/serenata-toolbox/tree/cuducos-lipedemorais-master) — feel free to take the commits if they help
* What's really missing is to address the residual files… this is still an issue, ok?

cuducos:
In the aforementioned branch I might have made things worse. Two residual changes after testing:

```
On branch cuducos-lipedemorais-master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   tests/fixtures/xz/federal-senate-reimbursements.xz

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	tests/fixtures/csv/federal-senate-2008.xz

no changes added to commit (use "git add" and/or "git commit -a")
```

tests/test_federal_senate_dataset.py
lipemorais: Done!

cuducos:
In 704f675 you removed some fixtures… are you shrinking the remaining ones to a couple of lines? The ones from my branch have 2 reimbursements each, pretty compact — feel free to grab them ; )

jtemporal:
>edit fixtures to have 2 lines each, for example (or more only if we have a plausible reason to add more)

I agree with @cuducos on this. If you want to test for all of the years reducing the fixtures to the minimum amount of lines in each would be the way to go IMHO

lipemorais:
@jtemporal @cuducos I completely agree, I removed some datasets that are not necessary anymore but there is space for improvement. I'm planing to fix this soon.

tests/test_federal_senate_dataset.py
lipemorais: I addressed it recently in this commit: 268148910768afb007b332d460e761dc4a12dfd6
What do you think of this approuch?

lipemorais:
> What's really missing is to address the residual files… this is still an issue, ok?

OK. I working on it.

I addressed the name convention here 0dfb560cae880b836d15b926712e1640747f4390


> In 704f675 you removed some fixtures… are you shrinking the remaining ones to a couple of lines? The ones from my branch have 2 reimbursements each, pretty compact — feel free to grab them ; )

I will grab just the idea, because the encoding is different. ;)

lipemorais:
@cuducos and @jtemporal I tunned the fixtures as much as I could to make it very fast eliminating generated files after somo tests that needed it.

Do you see any space for improvements?

cuducos:
Hell yeah! Many thanks @lipemorais for that improvements and @jtemporal and @anaschwendler for the support and code reviews  🎉

cuducos:
> I believe now it is only missing a version bump as @anaschwendler kindly reminded me earlier =)

Done.

serenata_toolbox/federal_senate/federal_senate_dataset.py
```diff
@@ -1,56 +1,95 @@
...
+        retrieved_files = []
+        not_found_files = []
+
+        for year in self.year_range:
             url = self.URL.format(year)
             file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
-            urlretrieve(url, file_path)
+            try:
+                urlretrieve(url, file_path)
+            except Exception as exception:
```
Irio: Trying to understand this try/catch during a debug session I'm going through at this moment. What is the reasoning behind catching an exception when one of the files is not found? Shouldn't an exception being raised something wanted in this case?

If that's not the case, at least catching something more specific like `FileNotFoundError` and not `Exception`.

lipemorais: That is a good point! 
I'm not sure if this is a all or nothing situation. I believe that you have more context than me for this.

> If that's not the case, at least catching something more specific like FileNotFoundError and not Exception.

Yeah, it should be more clear handling the exception. `FileNotFoundError` looks perfect for this.

PR #70, opened 2017-05-23
Add new companies dataset

cuducos:
New data, as the previous was not so good, as [discussed here](https://github.com/datasciencebr/serenata-de-amor/issues/214).

### DIFF
--- a/serenata_toolbox/datasets/downloader.py
+++ b/serenata_toolbox/datasets/downloader.py
@@ -37,7 +37,6 @@ class Downloader:
         '2017-02-15-receipts-texts.xz',
         '2017-03-15-reimbursements.xz',
         '2017-03-20-purchase-suppliers.xz',
-        '2017-04-27-companies-no-geolocation.xz',
         '2017-05-10-tse-candidates.xz',
         '2017-05-21-companies-no-geolocation.xz'
     )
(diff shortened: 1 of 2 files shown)


serenata_toolbox/datasets/downloader.py
```diff
@@ -38,7 +38,8 @@ class Downloader:
         '2017-03-15-reimbursements.xz',
         '2017-03-20-purchase-suppliers.xz',
         '2017-04-27-companies-no-geolocation.xz',
```
jtemporal: Shouldn't this be removed from this list in light of the new `companies-no-geolocation` dataset?

cuducos: I'm not sure. If we think about versioning, this should be there. Maybe someone somewhere in the world is running an analysis using it. But I do agree it's a quiet odd possibility. So… being strict, no this should not be deleted… but being pragmatic, yes, let's get rid of that.

jtemporal: I'm just saying the method is called `fetch_latest_backup` IMHO there should only be listed the most recent datasets. Am I wrong on assuming that? Maybe we should come up with a list of all the datasets available on the S3?

cuducos: Make sense then!

cuducos: Done in bbbf88b

Issue #75, opened 2017-05-24
Better config Coveralls for accuracy


othman853:
Hello, I hope it is not a problem to open an issue out of the blue like this. 

Lately, @lipemorais was showing me this project and, as a way to try to start contributing to it, I thought it would be interesting to replace Coveralls by another quality measuring tool. My reasons behind such suggestion are in the fact that the **Coveralls UI is very confusing**, I can't quickly get a picture of what is actually going on in the project when looking at its dashboard. Besides, **it is showing a quite imprecise metric** at this moment, since it's also considering third-party libraries in its analysis, lowering the metrics to only 25%. 

I believe that [CodeClimate](https://codeclimate.com) would be a better option to solve this problem. I've forked this project and configured its forked copy to be analyzed by CodeClimate. Below, it's possible to see a result of the first analysis:

<img width="1535" alt="screen shot 2017-05-24 at 12 57 19 am" src="https://cloud.githubusercontent.com/assets/4705127/26386310/4f9a0b2a-401c-11e7-9f98-c59501a35048.png">

Besides the code styling, it's also possible to harvest some metrics about testing coverage:

<img width="1679" alt="screen shot 2017-05-24 at 2 25 43 am" src="https://cloud.githubusercontent.com/assets/4705127/26388057/7fdf4dde-4028-11e7-8051-f0943252ede5.png">

<img width="1679" alt="screen shot 2017-05-24 at 2 25 59 am" src="https://cloud.githubusercontent.com/assets/4705127/26388098/b5028d96-4028-11e7-9d10-04fcd92dddbb.png">

I've also noticed that [Jarbas](https://github.com/datasciencebr/jarbas) makes use of both Coveralls and CodeClimate. Maybe, CodeClimate could be configured to be used as a test coverage reporter, replacing Coveralls there as well. 

I would like to apologize in advance if this is not the right way to open a new issue.

cuducos:
Hi @othman853,

Many thanks for you suggestion. Unfortunately I tend to disagree.

>  Coveralls UI is very confusing

This is a quite subjective matter — I use it a lot so I feel familiar in their dashboard.

> it is showing a quite imprecise metric at this moment, since it's also considering third-party libraries

[...]

PR #76, opened 2017-05-24
Ignoring virtualenv from coverage analysis

othman853:
This PR fixes the issue #75, by adding a coverage configuration file to fix the Coveralls inaccurate metrics (due to the unnecessary analysis of third-party libraries). Below, there is a screenshot showing the Coveralls statistics dashboard after applying the aforementioned configurations in the forked project:

screen shot 2017-05-24 at 2 09 01 pm

It is also possible to check this very same dashboard [here](https://coveralls.io/github/othman853/serenata-toolbox).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,2 @@
 [run]
   source = serenata_toolbox
-  omit =
-      */virtualenv/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+  source = serenata_toolbox
+  omit =
+      */virtualenv/*


.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
+  source = serenata_toolbox
+  omit =
```
cuducos: I don't think we need this `omit` (and the following line).

Once we setup `source = serenata_toolbox` the only think `coverage` will take into account is the `serenata_toobox` module (aka the directory `serenata_toolbox` in the root of the repo).

Besides that each user has a different name their for virtualenv (`~/.virtualenvs/` for those using the default of `virtualenvwrapper`, `.venv`, `venv`, `env`, `.env` etc…).

Any way, would you mind testing to check what reslts the `.coveragerc` without the `omit` gives us?

othman853: Sure! I can do this right away! I would just like to quickly remind that the `*/virtualenv/*` rule was used because of the libraries included by Travis during the build process, which I believe to be the source of the problem:

screen shot 2017-05-24 at 2 48 59 pm

This is slightly irrelevant for this issue's solution since the `source` section of the config file seems to be all we need to fix the problem (I'll update this PR accordingly as soon as all the tests are finished).

cuducos: >  I would just like to quickly remind that the `*/virtualenv/*` rule was used because of the libraries included by Travis during the build process

Wow! Haven't noticed it yet — really good point!

othman853: After removing the `omit` section, the results remained the same:

screen shot 2017-05-24 at 3 13 54 pm

I guess it is not necessary to keep such section by now. 😃

cuducos:
Yay! Many thanks for coding it and testing it ; )

PR #80, opened 2017-06-03
Add cities dataset to latest datasets download list

### LINKED ISSUE #78: Update Brazilian Cities CSV on the latest_backup list.
We are expanding our project to cities, in order to it we must update our dataset list.

We need to add the `2017-05-22-brazilian-cities.csv` to the `latest-backup` script, in order to make the `serenata-de_amor` notebooks work.

cuducos:
Fix #78

### DIFF
--- a/serenata_toolbox/datasets/downloader.py
+++ b/serenata_toolbox/datasets/downloader.py
@@ -39,6 +39,7 @@ class Downloader:
         '2017-03-20-purchase-suppliers.xz',
         '2017-05-10-tse-candidates.xz',
         '2017-05-21-companies-no-geolocation.xz',
+        '2017-05-22-brazilian-cities.csv',
         '2017-05-23-congressperson-details.xz'
     )
 
--- a/setup.py
+++ b/setup.py
@@ -33,5 +33,5 @@ setup(
         'serenata_toolbox.datasets'
     ],
     url=REPO_URL,
-    version='10.0.5'
+    version='10.0.6'
 )


cuducos:
```
def TestLatest(TestCase):

    def test_cities_was_added(self):
        self.assertIn('2017-05-22-brazilian-cities.csv', Datasets().downloader.LATEST)
```

Or even:

```
    def test_latest(self):
        self.assertEqual(28, len(Datasets().downloader.LATEST))
```

But the real question is would it worth it?

lipemorais:
I don't think that this kind of year worth.

What can be done to protect us from the future ourselves ?

May it should break if any of this files are removed from this list?

lipemorais:
... this kind of test* ...

cuducos:
> May it should break if any of this files are removed from this list?

I don't think so, because we remove old versions when there are new ones available… although we wold never see the `len()` of the list get smaller, because we would either replace a file (adding a newer version) or add a new dataset (which was not present in the list before).

lipemorais:
Why is this constant important for toolbox?

The LATEST constant is not used anywhere in this file, if I remove it all project tests will pass smoothly. 
Is it OK?

cuducos:
It's [used here](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/datasets/__init__.py#L71-L800), in a function [documented here](https://github.com/datasciencebr/serenata-toolbox#example-2-using-shortcuts).

Maybe the test you'd like to enhance is [this one](https://github.com/datasciencebr/serenata-toolbox/blob/master/tests/test_datasets.py#L82-L850).

lipemorais:
There is any point against enhance this tests?

cuducos:
> Is there any point against enhance this tests?

Not at all ; )

PR #81, opened 2017-06-03
Add a mini contribution guide

cuducos:
I dare anyone to _not_ find typos in this new `README.rst`

PR #83, opened 2017-06-05
Make integration tests run on CI in parallel with unit tests and caching dependencies

lipemorais:
I already tested it on Travis CI and worked as expected, here is the [builds on Travis](https://travis-ci.org/lipemorais/serenata-toolbox/builds).

The purpose of this PR is make the journey tests run on CI in parallel.

I moved the tests for journey and unit folders under tests.
I change .travis.yml file to run the tests in parallel and cache dependencies.
I removed the verification to check env variable before run integration tests

`python -m unittest tests/unit`
`python -m unittest tests/journey` 

@cuducos @jtemporal @anaschwendler @Irio


cuducos:
Wow! Pretty clever — very good indeed.

I'm just wondering… the journey test is kind of hidden, specially for newcomers (as integration tests envvar were before). Thinking of #81 is there a single command to be run that ends up running both unit and journey tests? `python -m unittest discover tests` runs only 44 tests, so I assumed its unit tests only… am I getting it wrong?

lipemorais:
@cuducos It happens because there is not __init__.py file inside journey folder. It's kind off good because CEAP journey test takes around **40 minutes**(It happened in my Linux at home yesterday) to execute itself.

lipemorais:
I just executed all the tests here and it took 1703.547s, almost 30 minutes.

cuducos:
I know it take ages… but… tests exist for a reason. I think people should run ir before opening PRs… that's why I suggested to add it to the default runner (and reinforce people can always use `tests/unit` for a quick check).

Issue #87, opened 2017-06-07
Big refactor of the public API to generate the datasets


cuducos:
This issue is proposed as a roadmap to a big refactor in the public API. This issue might also work as a _whishlist_ for you who uses this toolbox and believe its API for generating the datasets could be improved. I'll suggest a to-do list in this opening post and try to keep it updated as the following discussion goes by.

The main problems with the current one has been discussed by @lipemorais and myself in several other issues and PRs. For example:

* Problems with semantic: we only have `fetch`, `translate` and `clean` methods when what really happens in these three methods is fetch data, translate data to en_US, clean data, convert data from `.csv` to `.xz`, and merge datasets by year in a single file (see #53 and #68)
* These methods have a lot of logic embedded which makes tests more complex than they could be (see #68)
* We miss a single method to handle all tasks required to generate a dataste from the scratch (see #59)
* Integration tests still depends on external server — we could work with fixtures instead (also #59)

Therefore what I propose here is to:

- [ ] map the impact of changing the API
- [ ] rewrite `fetch`, `translate`, `clean` into more atomic methods ([reduce side effects](https://softwareengineering.stackexchange.com/questions/15269/why-are-side-effects-considered-evil-in-functional-programming)), with really simple logic and adding more methods if needed (e.g. `convert_to_xz`, `translate_to_en` etc.)
- [ ] add a method (e.g. `generate`) to handle all internal tasks from downloading it from the original source to having a dataset ready for Serenata pipeline (i.e. make all methods from the previous task internal methods used by this main one here)
- [ ] write unit tests for each of this methods that do not depend on external download (using mocks)
- [ ] write integration tests for this class that do not download on external download (using fixtures)

I think that this refactor will enhance our code quality and architecture and can pave the way to more overarching changes such as:

* adopting Dask as a default to handle any dataset (if we have barely no side effects, this is quite easy)
* changing the tests suite for something more robust such as `pytest`, or even using `tox`
* opening the API for new user customization, embracing DRY and avoiding [this](https://github.com/datasciencebr/whistleblower/blob/master/whistleblower/get_dataset.py)

JoaoCarabetta:
Hey,

is someone working on it? If positive, let me know!

I think that this toolbox should aim to cover all api endpoints from the Senate and Chamber of Deputies and other congress related data. I know that it seems a quite big dream. But, with that in mind, we should think on the structure of the project to easily accept new data entries in an organized way.

[...]

Issue #59, opened 2017-05-16
Simplify public API


cuducos:
IMHO our public API is quite redundant — surely redundancy can be good sometimes… so I ask does it worth it to simplify it in these terms?

Turns stuff like:

```python
from serenata_toolbox.federal_senate.federal_senate_dataset import FederalSenateDataset
```

In:

```python
from serenata_toolbox.federal_senate.dataset import Dataset
```

Or, at least"

```python
from serenata_toolbox.federal_senate.dataset import FederalSenateDataset
```

(And the same for `serenata_toolbox.chamber_of_deputies`).

anaschwendler:
So you think that we should work on it?
The only thing that I can say about that is that by doing that we will need to fix that in a few places, but it's ok to change that :)

@jtemporal thoughts?

lipemorais:
If were are talking about breaking changes...

What do you all think about the idea of have just one method to have a dataset, instead of call 3 methods(`fetch`, `clean` , `translate`)? I think that it will make the use of toolbox simpler and harder to break the API again.

anaschwendler:
For me it looks real great, me and @jtemporal could totally work on it. I will only wait for her approval and attempt to do that.

So we are stating that we are to refactor @cuducos and @lipemorais suggestions, what do you think @jtemporal?

jtemporal:
Hi y'all, sorry it took me a while to get to this issue.

As far as the fix the redundancy I believe is a great thing to work now. I would tackle it first and then consider reorganizing `fetch`, `clean`, `translate`.

Does that sound like a plan?

cuducos:
> What do you all think about the idea of have just one method to have a dataset, instead of call 3 methods(fetch, clean , translate)? 

I think it's a good idea, but it's **not** the focus of this issue. This issue was supposed to be a small refactor renaming some modules and classes — simple as that.

I reinforce that because I'm have been criticizing this whole public API put forward by the toolbox and I would like see a big refactor (as I said elsewhere, the `translate` method doesn't only translate etc.) and I think this is not the right place to handle it. I'll open an issue to discuss this overarching changes in the API.

lipemorais:
@cuducos I agree with you.

We can use baby step and have enhancements delivered in small packages.

anaschwendler:
@jtemporal @lipemorais @cuducos 
> This issue was supposed to be a small refactor renaming some modules and classes — simple as that.
[...]

PR #89, opened 2017-06-08
Simplify public API

anaschwendler:
This PR is related to #59.

So here, I refactor the names of the modules to simple ones.
In that journey, I noticed that we've completely forgot to check some files while working on refactoring the toolbox to accept both Chamber of Deputies and Federal Senate.

image

As we see in the image, it still keeps the format from before Federal Senate introduction  👀 
But I don't know if we just wait for the #87 to fix it, or something.

Help will be accepted <3

### DIFF
--- a/docs/source/modules.rst
+++ b/docs/source/modules.rst
@@ -1,0 +1,7 @@
+serenata_toolbox
+================
+
+.. toctree::
+   :maxdepth: 4
+
+   serenata_toolbox
--- a/docs/source/serenata_toolbox.chamber_of_deputies.rst
+++ b/docs/source/serenata_toolbox.chamber_of_deputies.rst
@@ -1,0 +1,62 @@
+serenata_toolbox.chamber_of_deputies package
+============================================
+
+Submodules
+----------
+
+serenata_toolbox.chamber_of_deputies.dataset module
+---------------------------------------------------
+
+.. automodule:: serenata_toolbox.chamber_of_deputies.dataset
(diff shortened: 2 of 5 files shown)


anaschwendler:
Ok, I think that I find a way to update the sphinx documentation.
What I've done was:
`make rst` into the `/docs` folder

So, I am not sure if it was the right way, I'll be waiting for reviews :)

docs/source/serenata_toolbox.rst
```diff
@@ -1,41 +1,14 @@
 serenata_toolbox package
 ========================
 
-Submodules
-----------
+Subpackages
```
cuducos: AFAIK they are submodules, not subpackages… but I assume this is merely 💅 So the real question is: have you changed it yourself or was it changed automagically? If it was automagically just let it be ; )

anaschwendler: Changed automagically! 
So, I can change it manually, but it was generated, so I don't know.

anaschwendler: I just run the makefile and 🎉

cuducos: In `sphinx` we trust then.

anaschwendler:
Ok, I guess that now everything is right!

cuducos:
> Ok, I guess that now everything is right!

Running integration tests and merge it as soon as it finishes ; )

PR #91, opened 2017-06-12
Introduces pytest to run all tests

lipemorais:

The purpose of this PR is make tests run with a more robust test runner to address part of #87 issue.

- Was added a configuration file to run test with pytest
- Was changed the command to run tests on Travis
- Added pytest as dependency

To test it just have coveralls installed `python -m pip install coveralls` and run the code below:
```
coverage run -m pytest tests/unit
coverage run -m pytest tests/journey
```
Who can help reviewing it?
@anaschwendler @cuducos @jtemporal


setup.py
```diff
@@ -21,7 +21,8 @@
         'beautifulsoup4>=4.4',
         'lxml>=3.6',
         'pandas>=0.18',
-        'tqdm'
+        'tqdm',
+        'pytest>3.1'
```
cuducos: Can you add id before `tqdm` to keep AZ ordering?

lipemorais:
Sure! I just reordered it. 

I also added the `travis_wait` before the command to run the tests. It should fix the CI.

.travis.yml
```diff
@@ -9,5 +9,5 @@ python: 3.5
 install:
   - python setup.py develop
   - pip install coveralls
-script: coverage run -m unittest discover tests/$TEST_SUITE
+script: travis_wait coverage run -m pytest tests/$TEST_SUITE
```
cuducos: [20min](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) sounds like not enough as tests CI failed…

lipemorais:
Now Travis CI Pipeline is working. :+1:

setup.py
```diff
@@ -21,6 +21,7 @@
         'beautifulsoup4>=4.4',
         'lxml>=3.6',
         'pandas>=0.18',
+        'pytest>3.1',
```
cuducos: Hum… `pytest` is not a install requirement (`install_requires`), it's a test requirement (`tests_requires`) according to [the docs](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=tests_require#new-and-changed-setup-keywords). This means two things:

1. `pytest` should not be installed for regular users, just for developers
2. We should run tests with `python setup.py test` if we're injecting tests dependencies

Does that make sense?

lipemorais: I completely agree. Thank you for the feedback.

I will make this improvement.

cuducos:
I couldn't test it yet but following the last commits I'd add two questions:

* If we're changing the way we test the package we need to update `README.rst`
* With this new implementation do we have to document different tests commands (eg how to run only unit tests instead f the whole suite)?

lipemorais:
> If we're changing the way we test the package we need to update README.rst

Sure! I updated README showing how to run tests using setup.py

> With this new implementation do we have to document different tests commands (eg how to run only unit tests instead f the whole suite)?

Nice! I added how to run unit and  journey tests separated on README

lipemorais:
@cuducos is there any change that have to be done to make this PR _mergeable_?

cuducos:
> @cuducos is there any change that have to be done to make this PR mergeable?

Gonna make a new _code review_, starting now. I spotted some possible changes I'd like to test before merging…

.travis.yml
```diff
@@ -2,12 +2,9 @@ language: python
...
     - $HOME/virtualenv/python3.5.3/
-env:
-  - TEST_SUITE=unit
-  - TEST_SUITE=journey
 python: 3.5
 install:
   - python setup.py develop
   - pip install coveralls
-script: coverage run -m unittest discover tests/$TEST_SUITE
+script: travis_wait 40 coverage run ./setup.py test
```
cuducos: No need for the `./` here.  `coverage run setup.py test` is enough ; )

README.rst
```diff
@@ -110,14 +110,21 @@ Always add tests to your contribution — if you want to test it locally before
 
 ::
 
-  $ python -m unittest discover tests
+  # Run all tests
+  $ python setup.py test
+
+  # Run all unit tests
+  $ pytest tests/unit
```
cuducos: `pytest` is only installed for tests, so this is gonna fail. Have you tried an envvar approach to separate unit test from integration tests? Something like: `python setup.py test` for unit tests and `RUN_INTEGRATION_TESTS=1 python setup.py test` for integration tests?

README.rst
```diff
@@ -110,14 +110,21 @@ Always add tests to your contribution — if you want to test it locally before
...
+  # Run all journey tests
+  $ pytest tests/journey
 
 When the tests are passing, also check for coverage of the modules you edited or added — if you want to check it before opening the PR:
 
 ::
 
   $ pip install coverage
-  $ coverage run -m unittest discover tests
+  $ coverage run ./setup.py test
```
cuducos: No need for the `./` here.  `coverage run setup.py test` is enough ; )

lipemorais:
> I added some inline comments — we cannot rely on pytest command when pytest is installed for tests only

We will need to add the instructions to install it with pip, as we do with isort and prospector.
Fixed here 37444b900e049e928e3b57d9e6587e6aa715f6cf

> Besides that I couldn't figure out why there are so many changes in test files themselves in this PR. Can you clarify (or, eventually, cleanup and leave this for a new PR)?

Some unit tests were not been executed because was not under `tests/unit` folder, so I moved them to the unit folder.

> No need for the ./ here.  coverage run setup.py test is enough ; )

Thanks! Fixed in 1ce5a99aa9f42cf957744db75a67a62f98969876.

> Have you tried an envvar approach to separate unit test from integration tests? Something like: python setup.py test for unit tests and RUN_INTEGRATION_TESTS=1 python setup.py test for integration tests?

It seams a little confusing because the second command,` RUN_INTEGRATION_TESTS=1 python setup.py test`  will be no able to run just integrations tests, it will run unit + integration. I think that is not simple and explicit. I feel that with an approach like this we will walk back in the timeline when the integration tests where not running on CI. 

What do you think @cuducos ?

lipemorais:
@cuducos @jtemporal @Irio @anaschwendler 

Is there any point that should improved to make this PR able to be merged that was not addressed?

cuducos:
I must say that in general this PR is missing its purpose:

> The purpose of this PR is make tests run with a more robust test runner to address part of #87 issue.

IMHO using `pytest` only as a short way to call tests has nothing to do with #87. The complexity mentioned in that issue is related to architecture, not with test runner. Problematic architecture (package architecture or tests architecture—or even both, which I think is the case) will result in unorganized and complex code whether it is with `unittest` or `pytest`.

Surely I mentioned `pytest` in #87:

> I think that this refactor will enhance our code quality and architecture and can pave the way to more overarching changes such as:
>
> * [ …]
> * changing the tests suite for something more robust such as `pytest`, or even using `tox`

So I truly believe changing the test runner/suite should come in a better moment, that is to say: after refactoring the toolbox, not as a way to refactor it.

---

That said I'll leave it to another code reviewer, but my vote is to close this issue and address marginal pros it has in another, atomic PRs (for example: run integration and unit tests in parallel in CI, remove of duplicate logic, code style and clean up).

PR #96, opened 2017-06-17
Remove dtype for cols already with converters

### LINKED ISSUE #95: "Both a converter and dtype were specified" warning with chamber_of_deputies
When using `pandas.read_csv`, you must provide a converter or a dtype, not both.

**Log**
```console
$ python rosie.py run chamber_of_deputies
/home/travis/virtualenv/python3.6.1/lib/python3.6/site-packages/serenata_toolbox/chamber_of_deputies/dataset.py:74: ParserWarning: Both a converter and dtype were specified for column vlrDocumento - only the converter will be used
  'vlrRestituicao': lambda x: float(x.replace(',','.'))})

/home/travis/virtualenv/python3.6.1/lib/python3.6/site-packages/serenata_toolbox/chamber_of_deputies/dataset.py:74: ParserWarning: Both a converter and dtype were specified for column vlrGlosa - only the converter will be used
  'vlrRestituicao': lambda x: float(x.replace(',','.'))})

/home/travis/virtualenv/python3.6.1/lib/python3.6/site-packages/serenata_toolbox/chamber_of_deputies/dataset.py:74: ParserWarning: Both a converter and dtype were specified for column vlrLiquido - only the converter will be used
  'vlrRestituicao': lambda x: float(x.replace(',','.'))})

/home/travis/virtualenv/python3.6.1/lib/python3.6/site-packages/serenata_toolbox/chamber_of_deputies/dataset.py:74: ParserWarning: Both a converter and dtype were specified for column vlrRestituicao - only the converter will be used
  'vlrRestituicao': lambda x: float(x.replace(',','.'))})

/home/travis/build/datasciencebr/rosie/rosie/chamber_of_deputies/adapter.py:25: DtypeWarning: Columns (10) have mixed types. Specify dtype option on import or set low_memory=False.
  self.update_datasets()

Merging all datasets…
Loading reimbursements-2009.xz…
Loading reimbursements-2010.xz…
Loading reimbursements-2011.xz…
Loading reimbursements-2012.xz…
Loading reimbursements-2013.xz…
Loading reimbursements-2014.xz…
Loading reimbursements-2015.xz…
Loading reimbursements-2016.xz…
Loading reimbursements-2017.xz…
Dropping rows without document_value or reimbursement_number…
Grouping dataset by applicant_id, document_id and year…
Gathering all reimbursement numbers together…
Summing all net values together…
Summing all reimbursement values together…
Generating the new dataset…
Casting changes to a new DataFrame…
Writing it to file…
Done.
```
https://travis-ci.org/datasciencebr/rosie/builds/244041874

Must be changed in
https://github.com/datasciencebr/serenata-toolbox/blob/bced7b332c486f6a20f5bbc1285cf843040c76ed/serenata_toolbox/chamber_of_deputies/dataset.py#L56-L74

Irio:
Getting rid of a warning.

Fixes #95.


Irio:
For the one merging this: can you squash and merge it?

serenata_toolbox/chamber_of_deputies/dataset.py
```diff
@@ -137,3 +132,6 @@ def __translate_file(self, csv_path):
                     encoding='utf-8')
 
         return output_file_path
+
+    def __parse_float(self, string):
```
lipemorais: There is any special motive to have two trailing underscore here? 

Could we use just one to say that it's just to be used internally and not a special python method.

Irio: Just because preceding a method with a double underscore will make it inaccessible from outside of the object.

```
In [1]: class Object:
   ...:     def foo(self):
   ...:         return 1
   ...:     def _foo(self):
   ...:         return 2
   ...:     def __foo(self):
   ...:         return 3
   ...:

In [2]: Object().foo()
Out[2]: 1

In [3]: Object()._foo()
Out[3]: 2

In [4]: Object().__foo()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-b254b9c59e17> in <module>()
----> 1 Object().__foo()

AttributeError: 'Object' object has no attribute '__foo'
```

lipemorais: Usually in Python use kind of prevention is not necessary.

> “Private” instance variables that cannot be accessed except from inside an object don’t exist in Python. However, there is a convention that is followed by most Python code: a name prefixed with an underscore (e.g. _spam) should be treated as a non-public part of the API (whether it is a function, a method or a data member). It should be considered an implementation detail and subject to change without notice.

source: https://docs.python.org/3/tutorial/classes.html#private-variables

@Irio, do you feel that it still worth it?

Irio: In this case I agree with the change. But feel it should probably come in a new pull request, since this one does not intend to apply this pattern across the repository.

These are the results when searching for the given regex across the project:
`def _[a-z]` -> 5 results found
`def __(?!init)` -> 10 results found

lipemorais:
Nice! A issue can be open a new contributor could use it as first contribution to the project as easy help.

PR #97, opened 2017-06-18
Allow the user to provide a year list to filter the datasets

Irio:
One step in the direction to allow users to make non-standard uses of the datasets.

In the Whistleblower repository, will remove the need of copying and pasting a whole file - [whistleblower/get_dataset.py](https://github.com/datasciencebr/whistleblower/blob/25c4c8932228143e61b0f8eb6a764aeabbaffa7a/whistleblower/get_dataset.py) - for just a few changes.

### DIFF
--- a/serenata_toolbox/chamber_of_deputies/dataset.py
+++ b/serenata_toolbox/chamber_of_deputies/dataset.py
@@ -7,18 +7,19 @@ import numpy as np
 import pandas as pd
 from .reimbursements import Reimbursements
 
+
 class Dataset:
 
-    YEARS = [n for n in range(2009, date.today().year+1)]
+    AVAILABLE_YEARS = [year for year in range(2009, date.today().year + 1)]
 
-    def __init__(self, path):
+    def __init__(self, path, years=AVAILABLE_YEARS):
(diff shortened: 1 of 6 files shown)


serenata_toolbox/chamber_of_deputies/dataset.py
```diff
@@ -7,18 +7,19 @@
...
 class Dataset:
 
-    YEARS = [n for n in range(2009, date.today().year+1)]
+    AVAILABLE_YEARS = [year for year in range(2009, date.today().year + 1)]
 
-    def __init__(self, path):
+    def __init__(self, path, years=AVAILABLE_YEARS):
         self.path = path
-
+        self.years = years
```
cuducos: Wouldn't it be nice to have this argument flexible? Year could be an `int` (eg. `Dataset('/tmp`, years=2010)`) or a `list` (eg. `Dataset('/tmp`, years=[2010, 2011])`). In that case I'd replace this line by:

```python
        self.years = [years] if isinstance(year, int) else years
```

Irio:
Tests passed after the change. Rebasing with master.

Issue #98, opened 2017-06-18
Failing tests don't make the suite/CI fail


Irio:
Current status is a test suite failing with a green flag. When running locally, I get the same errors with a zero exit code.
https://travis-ci.org/datasciencebr/serenata-toolbox/jobs/242200049

Irio:
At least few of the warnings shown are given to missing mocks. Adding them should remove message logs from the suite and let us see the real output.

lipemorais:
Could share some of the logs here? I just can see a green build :/

cuducos:
> Current status is a test suite failing with a green flag

What tells you the tests are failing? Can you be more specific?

Warning messages about `config.ini` are actually expected in [tests like those](https://github.com/datasciencebr/serenata-toolbox/blob/master/tests/unit/test_datasets_remote.py#L10-L24).

cuducos:
> Could share some of the logs here? I just can see a green build :/

Logs are in Travis build @Irio posted.

Being green arguably is the bug if he's right about the expectation of red for this scenario — I just can't see why he expected this test to be red.

lipemorais:
> Logs are in Travis build @Irio posted.

Yeah, I got it. I'm not sure which is the warning mentioned that should breaks the build.

After look at a comment here #68 I think that the warning is: `While fetching Seranata Toolbox not found file: /tmp/federal-senate-2007.csv ` inside the Travis CI log mentioned by @Irio.

jtemporal:
since #105 was merged, I'm closing this one ;)

PR #105, opened 2017-07-02
Handle exceptions in a better way for federal_senate_dataset

lipemorais:

The purpose of this PR is handle exceptions as expected raised on #98 by @irio.

I changed the just logging and continue approach to raise the exceptions after logging about it.

Just run unit tests is enough to check it because some tests were added with the use cases. To be more specific.

```python -m unittest tests/unit/test_federal_senate_dataset.py```

@Irio was who raised the point, so I think that would be the best person to give feedback about it. Other people that can help are @anaschwendler  @jtemporal @cuducos. If you are not one of them feel free to provide me some feedbacks as well.

### DIFF
--- a/serenata_toolbox/federal_senate/dataset.py
+++ b/serenata_toolbox/federal_senate/dataset.py
@@ -1,7 +1,7 @@
 import os.path
 from datetime import date
+from urllib.error import HTTPError, URLError
 from urllib.request import urlretrieve
-from urllib.error import (HTTPError, URLError)
 
 import pandas as pd
 
--- a/serenata_toolbox/federal_senate/dataset.py
+++ b/serenata_toolbox/federal_senate/dataset.py
@@ -1,4 +1,5 @@
 import os.path
+import urllib
 from datetime import date
 from urllib.request import urlretrieve
 
@@ -23,9 +24,16 @@ class Dataset:
             file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
             try:
                 urlretrieve(url, file_path)
-            except Exception as exception:
-                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, exception))
-                not_found_files.append(file_path)
+            except urllib.error.URLError as url_error_exception:
+                print('The server couldn\'t fulfill the request.')
+                print('Reason: ', url_error_exception.reason)
+                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, url_error_exception))
+                raise url_error_exception
+            except urllib.error.HTTPError as http_erro_exception:
+                print('We failed to reach the server')
+                print('Reason:  ', url_error_exception.reason)
+                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, url_error_exception))
+                raise url_error_exception
             else:
                 retrieved_files.append(file_path)
 
@@ -40,9 +48,9 @@ class Dataset:
             csv_path = os.path.join(self.path, filename)
             try:
                 self._translate_file(csv_path)
-            except Exception as exception:
-                print('While translating Seranata Toolbox not found file: {} \n{}'.format(csv_path, exception))
-                not_found_files.append(csv_path)
+            except FileNotFoundError as file_not_found_error:
+                print('While translating Seranata Toolbox not found file: {} \n{}'.format(csv_path, file_not_found_error))
+                raise file_not_found_error
             else:
                 translated_files.append(csv_path)
 
--- a/serenata_toolbox/federal_senate/dataset.py
+++ b/serenata_toolbox/federal_senate/dataset.py
@@ -1,7 +1,7 @@
 import os.path
-import urllib
 from datetime import date
 from urllib.request import urlretrieve
+from urllib.error import (HTTPError, URLError)
 
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: serenata_toolbox/federal_senate/dataset.py [hunk 1]
```diff
@@ -25,9 +26,16 @@ def fetch(self):
...
-            except Exception as exception:
-                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, exception))
-                not_found_files.append(file_path)
+            except HTTPError as http_error_exception:
+                print('We failed to reach the server')
+                print('Error code  ', http_error_exception.reason)
+                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, http_error_exception))
+                raise http_error_exception
+            except URLError as url_error_exception:
+                print('The server couldn\'t fulfill the request.')
```
jtemporal: Following [PEP8 on string quotes](https://www.python.org/dev/peps/pep-0008/#string-quotes):
>When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string

So let's remove this nasty backslash 😬

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -25,9 +26,16 @@ def fetch(self):
             file_path = os.path.join(self.path, 'federal-senate-{}.csv'.format(year))
             try:
                 urlretrieve(url, file_path)
-            except Exception as exception:
-                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, exception))
-                not_found_files.append(file_path)
+            except HTTPError as http_error_exception:
+                print('We failed to reach the server')
+                print('Error code  ', http_error_exception.reason)
```
jtemporal: I believe there is an extra space there. You can delete it ;)

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -25,9 +26,16 @@ def fetch(self):
...
+                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, http_error_exception))
```
jtemporal: I would change the construction here to: `'While fetching, Serenata Toolbox didn't find the file: {} \n{}'`

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -25,9 +26,16 @@ def fetch(self):
...
+                raise http_error_exception
+            except URLError as url_error_exception:
+                print('The server couldn\'t fulfill the request.')
+                print('Reason: ', url_error_exception.reason)
+                print('While fetching Seranata Toolbox not found file: {} \n{}'.format(file_path, url_error_exception))
```
jtemporal: same as before: `'While fetching, Serenata Toolbox didn't find the file: {} \n{}'`

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -42,9 +50,9 @@ def translate(self):
             csv_path = os.path.join(self.path, filename)
             try:
                 self._translate_file(csv_path)
-            except Exception as exception:
-                print('While translating Seranata Toolbox not found file: {} \n{}'.format(csv_path, exception))
-                not_found_files.append(csv_path)
+            except FileNotFoundError as file_not_found_error:
+                print('While translating Seranata Toolbox not found file: {} \n{}'.format(csv_path, file_not_found_error))
```
jtemporal: Same as before: `'While translating, Serenata Toolbox didn't find the file: {} \n{}'`

tests/unit/test_federal_senate_dataset.py
```diff
@@ -15,88 +16,95 @@ def setUpClass(cls):
...
+                          2009)
 
-        file_path = os.path.join(self.subject.path, 'federal-senate-2008.csv')
+        file_path = os.path.join(subject.path, 'federal-senate-2008.csv')
         federal_senate_2008 = pd.read_csv(file_path,
                                           sep=';',
                                           encoding='ISO-8859-1',
                                           skiprows=1)
         self.assertIsNotNone(federal_senate_2008['ANO'],
                              'expects \'ANO\' as column in this dataset')
```
jtemporal: Following [PEP8 on string quotes](https://www.python.org/dev/peps/pep-0008/#string-quotes):
>When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string

So let's remove this nasty backslash 😬

tests/unit/test_federal_senate_dataset.py
```diff
@@ -15,88 +16,95 @@ def setUpClass(cls):
...
-        self.subject.translate()
+        subject.translate()
 
-        translated_file_path = os.path.join(self.subject.path, 'federal-senate-2008.xz')
+        translated_file_path = os.path.join(subject.path, 'federal-senate-2008.xz')
         translated_federal_senate_2008 = pd.read_csv(translated_file_path,
                                                      encoding='utf-8')
 
         self.assertIsNotNone(translated_federal_senate_2008['year'],
                              'expects \'year\' as column in this dataset')
```
jtemporal: Following [PEP8 on string quotes](https://www.python.org/dev/peps/pep-0008/#string-quotes):
>When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string

So let's remove this nasty backslash 😬

jtemporal:
@lipemorais some conflicts showed up, would mind fixing them up please?

lipemorais:
I made fix expecting to solve the conflicts @jtemporal :)

jtemporal:
This looks perfect! Only thing missing here is a version bump! I believe we should go with a `MINOR` @cuducos what do you think?

lipemorais:
I'm not sure since we are throwing exceptions now, isn't it a breaking change? What do you think @cuducos and @jtemporal ?

cuducos:
> I'm not sure since we are throwing exceptions now, isn't it a breaking change?

Indeed — I haven't thought about that before. IMHO it doesn't change the API but change the behavior of the application; so major bump.

lipemorais:
Bumped!

PR #113, opened 2017-07-20
Remove double underscore

### LINKED ISSUE #99: Replace methods name starting with double underscore (use a single underscore instead)
As [discussed here](https://github.com/datasciencebr/serenata-toolbox/pull/96#discussion_r122927316) — thanks @lipemorais and @Irio.

jtemporal:
This should fix #99 ;)


jtemporal:
I'm thinking this should be a MAJOR bump, @cuducos any thoughts?

cuducos:
`_methods` are supposed to be private, not part of the public API. IMHO it's just _micro_/_patch_: no change in the public API, no new features, no backward incompatibility…

jtemporal:
bumped! waiting for checks to pass =)

Issue #46, opened 2017-04-28
Amazon credentials path


luipillmann:
Error when running Jupyter notebooks in **serenata-de-amor/develop** saying it could not find **config.ini** file (Amazon credentials).

```
Could not find config.ini file.
using fetch from serenata_toolbox.datasets
`Could not find config.ini file.
You need Amzon section in it to interact with S3
(Check config.ini.example if you need a reference.)
—-------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-b724014e1ca7> in <module:angry:)
—--> 1 fetch('2017-04-21-sex-place-distances.xz', '../data')

/Users/renangohe/anaconda/envs/serenata_de_amor/lib/python3.6/site-packages/serenata_toolbox/datasets/__init__.py in fetch(filename, destination_path)
     72 
     73 def fetch(filename, destination_path):
—-> 74     datasets = Datasets(destination_path)
     75     return datasets.downloader.download(filename)
     76 

/Users/renangohe/anaconda/envs/serenata_de_amor/lib/python3.6/site-packages/serenata_toolbox/datasets/__init__.py in __init__(self, local_directory)
     53             local_directory,
     54             bucket=self.remote.bucket,
—-> 55             **self.remote.credentials
     56         )
     57 

TypeError: type object argument after ** must be a mapping, not NoneType
```
We are working on implementing a simple search for the config file in the project folder.

cuducos:
Hi @luipillmann,

Can you provide more detail abou the code you're running?

Also I'm not sure what you mean by _running Jupyter notebooks in serenata-de-amor/develop_: if you're running `jupyter notebook` from that directory that is not how we thought things would work. Run `jupyter notebook` from the root, not from `develop/` (or, if you really want to run from `develop/`, copy `config.ini` from root to there).

BTW do you have a `config.ini` in the root of the project?

luipillmann:
Hi, @cuducos. 

We were running `jupyter notebook` from the root, where the config.ini file was placed (its content was copied from the example .ini). 

What I meant is that the **.ipnb** files were run from `develop/`. It seems the program expects config.ini to be in the same folder where notebooks are. 

We suppose it has something to do with line 12 in [datasets/remote.py](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/datasets/remote.py)

cuducos:
Ow…got it. Makes sense. 

This should fix it:

```python
[...]

PR #114, opened 2017-07-21
add config file lookup feature

humrochagf:
Hey, i made this PR in order to fix issue #46.

I've added a lookup function inspired on [fabric's fabfile discovery](https://github.com/fabric/fabric/blob/master/fabric/main.py#L106-L117) to go down from current directory to root looking for a `config.ini` file. If the file is not found it returns the informed file name.

I leave a comment on that function talking about the possibility of returning None on a future refactor deprecating the `config_exists` remote attribute. But it probably need some checking on possible code that rely on that.

### DIFF
--- a/serenata_toolbox/datasets/helpers.py
+++ b/serenata_toolbox/datasets/helpers.py
@@ -72,7 +72,7 @@ def find_config(name='config.ini'):
         joined = os.path.join(path, name)
 
         # return config file absolute path if exists
-        if os.path.exists(joined):
+        if os.path.exists(joined) and os.path.isfile(joined):
             return os.path.abspath(joined)
 
         path = os.path.join('..', path)
--- a/serenata_toolbox/datasets/helpers.py
+++ b/serenata_toolbox/datasets/helpers.py
@@ -56,3 +56,30 @@ def save_to_csv(df, data_dir, name):
     today = datetime.strftime(datetime.now(), '%Y-%m-%d')
     file_path = os.path.join(data_dir, '{}-{}.xz'.format(today, name))
     df.to_csv(file_path, **CSV_PARAMS)
+
+
+# Utility for config file lookup (credits to fabric fabfile lookup)
+
+def find_config(name='config.ini'):
+    """
+    :param name: config file name (defaults to 'config.ini')
+    """
+    # start in cwd and work downwards towards filesystem root
+    path = '.'
+    # Stop before falling off root of filesystem (should be platform
+    # agnostic)
+    while os.path.split(os.path.abspath(path))[1]:
+        joined = os.path.join(path, name)
+
+        # return config file absolute path if exists
+        if os.path.exists(joined):
+            return os.path.abspath(joined)
+
+        path = os.path.join('..', path)
+
+    # if config not found fallback on tests expected behavior
+    # wich is to return the filename to be used at remote
+    # config_exists class property.
+    # a future refactor could be to return None and not use
+    # config exists anymore
+    return name
--- a/serenata_toolbox/datasets/remote.py
+++ b/serenata_toolbox/datasets/remote.py
@@ -5,6 +5,7 @@ import os
 import boto3
 
 from serenata_toolbox.datasets.contextmanager import status_message
+from serenata_toolbox.datasets.helpers import find_config
 
 
 class RemoteDatasets:
@@ -14,6 +15,7 @@ class RemoteDatasets:
     def __init__(self):
         self.credentials = None
         self.client = None
+        self.config = find_config(self.CONFIG)
 
         if not self.config_exists:
             print('Could not find {} file.'.format(self.CONFIG))
@@ -22,7 +24,7 @@ class RemoteDatasets:
             return
 
         settings = configparser.RawConfigParser()
-        settings.read(self.CONFIG)
+        settings.read(self.config)
         self.settings = partial(settings.get, 'Amazon')
 
         try:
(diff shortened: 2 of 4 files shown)


cuducos:
Looks good to me! Thanks @humrochagf.

What about instead of returning `None`, returning the contents of `config.ini.example`? For most cases, just those keys are enough… fell free to add that here if you like the idea, otherwise we can open a new issue. Just LMK. 

What do you think, @jtemporal?

cuducos:
Ow, one ore thing: Any special reason for the `find_config`

cuducos:
Sorry, just pressed the wrong button ; )

jtemporal:
>What about instead of returning None, returning the contents of config.ini.example?

I like this idea.

humrochagf:
I would sugest keeping the idea of returning None and making a second helper function `load_config` encharged to load the file or loading a default configuration on None case.

It's more easily testable and split these two tasks in well defined blocks of code

jtemporal:
Nice PR @humrochagf this will help a lot. Even though is passing on our CI, the test is failing on my machine. Any thoughts? Here's the failing message and the traceback.

screen shot 2017-07-21 at 10 34 58 am


[Complete traceback](https://gist.github.com/jtemporal/9f3a5c47386c3c52af6c47631d2baf92)

ps.: I'm up for pairing to fix this ;)

jtemporal:
>I would sugest keeping the idea of returning None and making a second helper function load_config encharged to load the file or loading a default configuration on None case.

Agreed!

cabral:
Runs OK on UBUNTU 16 LTS

humrochagf:
added more tests to cover the case were confi.ini is a file and fixed the unexpected behavior and made a try on fixing the mac problem @jtemporal can you check it?

jtemporal:
@humrochagf no change, still failing =/

cuducos:
> I would sugest keeping the idea of returning None and making a second helper function `load_config` encharged to load the file or loading a default configuration on `None` case.

Couldn't agree more, @humrochagf!

About the error: 

```
- /private/var/folders/dj/wl7_2hgs0652svdpc6gxvmq40000gn/T/tmposhuzm_2/config.ini
? --------
+ /var/folders/dj/wl7_2hgs0652svdpc6gxvmq40000gn/T/tmposhuzm_2/config.ini
```

`/var` in macOS [seems to be](https://stackoverflow.com/questions/12482702/pythons-os-chdir-and-os-getcwd-mismatch-when-using-tempfile-mkdtemp-on-ma) a symlink to `/private/var` — so we have to make sure we workaround that for macOS so the intended comparison works. If I'm not wrong [`os.path.realpath`](https://docs.python.org/3/library/os.path.html#os.path.realpath) can be helpful here ; )

humrochagf:
Right, this last commit applies this approach, let's see if that fixes the problem.

r41278856:
Hi,
I am running rosie/rosie.py chamber_of_deputies in serenata-de-amor and I am facing a problem similar to [https://github.com/okfn-brasil/serenata-toolbox/issues/46](url).

`2018-07-18 13:34:02,000 - root - INFO - Could not find config.ini file.
2018-07-18 13:34:02,001 - root - INFO - You need Amazon section in it to interact with S3
2018-07-18 13:34:02,001 - root - INFO - (Check config.ini.example if you need a reference.)
Traceback (most recent call last):
  File "./rosie/rosie.py", line 60, in <module>
    command()
  File "./rosie/rosie.py", line 34, in run
    klass.main(target_directory)
  File "/Users/jimenacastillo/Documents/mres_dissertation/serenata-de-amor/rosie/rosie/chamber_of_deputies/__init__.py", line 8, in main
    core = Core(settings, adapter)
  File "/Users/jimenacastillo/Documents/mres_dissertation/serenata-de-amor/rosie/rosie/core/__init__.py", line 30, in init
    self.dataset = adapter.dataset
  File "/Users/jimenacastillo/Documents/mres_dissertation/serenata-de-amor/rosie/rosie/chamber_of_deputies/adapter.py", line 25, in dataset
    self.update_datasets()
  File "/Users/jimenacastillo/Documents/mres_dissertation/serenata-de-amor/rosie/rosie/chamber_of_deputies/adapter.py", line 70, in update_datasets
    fetch(self.COMPANIES_DATASET, self.path)
  File "/Users/jimenacastillo/miniconda3/envs/serenata_rosie/lib/python3.6/site-packages/serenata_toolbox/datasets/__init__.py", line 78, in fetch
    datasets = Datasets(destination_path)
  File "/Users/jimenacastillo/miniconda3/envs/serenata_rosie/lib/python3.6/site-packages/serenata_toolbox/datasets/__init__.py", line 59, in init
    **self.remote.credentials
TypeError: type object argument after ** must be a mapping, not NoneType`

I using conda in a Mac and I am very beginner in programming.
Hope someone could help me, thanks!

@cuducos

cuducos:
@r41278856, can you also share the command you entered to output this error?

r41278856:
Yes, sure
python ./rosie/rosie.py run chamber_of_deputies /Users/jimenacastillo/Documents/mres_dissertation/data

r41278856:
It was prolemb with the path. I solved it. Thanks

cuducos:
> It was prolemb with the path. I solved it.

Can you share how you solved? It might help other users facing the same issue/error ; )

r41278856:
Sure, the problem. Look at the change of directories that should be made to run the files.

cd serenata-de-amor

conda create -n  serenata_rosie python=3

source activate serenata_rosie

conda install pip=9.0.1

./rosie/setup

conda install pip=10.0.1

cd rosie

python ./rosie.py run chamber_of_deputies

PR #115, opened 2017-07-21
Add Semantic Versioning link

cuducos:


jtemporal:
Journey tests a failing because the Federal Senate's website is down. I'm merging this inspire of that.

PR #116, opened 2017-07-25
Adds Operação Serenata de Amor e-mail to setup file

jtemporal:
This should fix the warning generated while running the `sdist` command for PyPI release. 

`warning: check: missing meta-data: if 'author' supplied, 'author_email' must be supplied too`


jtemporal:
I'm not sure but I don't think this requires a version bump. Although this could be considered a patch.

lipemorais:
I believe that just a patch bump is enough, @jtemporal.

PR #119, opened 2017-07-27
Updates toolbox instalation instructions

### LINKED ISSUE #118: Documentation needs update
Installation now can be done via `pip` it can be updated

jtemporal:
fix #118

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,7 @@ Installation
 
 ::
 
-    $ pip install git+https://github.com/datasciencebr/serenata-toolbox.git#egg=serenata-toolbox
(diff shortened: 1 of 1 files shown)


jtemporal:
> Since the project uses Python 3. Shouldn't we use python -m pip install serenata-toolbox?
What do you think @jtemporal ?

@lipemorais I think that's totally "doable" but it should be in another issue. Also would open up the issue on the other repos to do the same 😉

Right now the focus is only to show that serenata-toolbox is only a `pip install` away.

@cuducos any thoughts?

lipemorais:
I completely agree @jtemporal

PR #121, opened 2017-07-28
Updates latest file list

### LINKED ISSUE #120: Document new reimbursement dataset available
There is a more recent reimbursements dataset available on our S3 but it isn't documented yet.

`2017-07-04-reimbursements.xz` needs to be added to our `LATEST` on `downloader.py`

jtemporal:
fix #120


lipemorais:
One more file to download is making the journey test even taking more time. :(
We need to address it as soon as possible.

PR #129, opened 2017-08-11
Adds condition to deploy

### LINKED ISSUE #127: Deploy only after Unit tests passes
Right now Travis tries to deploy to PyPI two time hence breaking the build in the second time.

 Maybe add a condition for the deploy only to be done if unit tests pass would be a smart choice.

jtemporal:
fix #127 This should make the deploy only happen once.


jtemporal:
I'm closing this one in favor of #134 ;)

PR #132, opened 2017-08-19
Filter missing datasets for download

tuliocasagrande:
During the projects' setup, the toolbox downloads dozens of datasets. If a connection error occurs during the downloads and the user tries again, the toolbox is going to download all datasets, even those that were successfully downloaded previously.

This PR filters for missing datasets to be downloaded.
Future work can include a `force` parameter to ignore and download every dataset regardless of their presence.


serenata_toolbox/datasets/__init__.py
```diff
@@ -74,7 +74,12 @@ def fetch(filename, destination_path):
...
     return datasets.downloader.download(filename)
 
+
 def fetch_latest_backup(destination_path):
     datasets = Datasets(destination_path)
     files = datasets.downloader.LATEST
+
+    files = tuple(filter(lambda f: not os.path.exists(
+        os.path.join(destination_path, f)), files))
+
```
cuducos: What about that? Does it sound a little bit more pythonic?

```python
files = tuple(
    f for f in datasets.downloader.LATEST
    if not os.path.exists(destination_path, f)
)
```

tuliocasagrande: 👍

cuducos:
This PR is red but [the error](https://travis-ci.org/datasciencebr/serenata-toolbox/builds/266316452) doesn't seem to be related to changes from this PR. Any ideas @jtemporal @lipemorais? Maybe the Senado servers are serving something unexpected?

<details>

```
======================================================================
ERROR: test_journey_federal_senate_dataset (test_federal_senate_dataset.TestJourneyFederalSenateDataset)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/datasciencebr/serenata-toolbox/tests/journey/test_federal_senate_dataset.py", line 22, in test_journey_federal_senate_dataset
    self.subject.translate()
  File "/home/travis/build/datasciencebr/serenata-toolbox/serenata_toolbox/federal_senate/dataset.py", line 50, in translate
  File "/home/travis/build/datasciencebr/serenata-toolbox/serenata_toolbox/federal_senate/dataset.py", line 99, in _translate_file
    skiprows=1)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/io/parsers.py", line 655, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/io/parsers.py", line 405, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/io/parsers.py", line 764, in __init__
    self._make_engine(self.engine)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/io/parsers.py", line 985, in _make_engine
    self._engine = CParserWrapper(self.f, **self.options)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/pandas-0.20.3-py3.5-linux-x86_64.egg/pandas/io/parsers.py", line 1605, in __init__
    self._reader = parsers.TextReader(src, **kwds)
  File "pandas/_libs/parsers.pyx", line 565, in pandas._libs.parsers.TextReader.__cinit__ (pandas/_libs/parsers.c:6260)
pandas.errors.EmptyDataError: No columns to parse from file
----------------------------------------------------------------------
Ran 7 tests in 802.064s
FAILED (errors=1)
```

</details>

jtemporal:
>Maybe the Senado servers are serving something unexpected?

I Restarted the build to see if the error persists. If does I'll take a closer look into it.

serenata_toolbox/datasets/__init__.py
```diff
@@ -74,7 +74,13 @@ def fetch(filename, destination_path):
...
     return datasets.downloader.download(filename)
 
+
 def fetch_latest_backup(destination_path):
     datasets = Datasets(destination_path)
-    files = datasets.downloader.LATEST
+
+    files = tuple(
+        f for f in datasets.downloader.LATEST
+        if not os.path.exists(destination_path, f)
```
jtemporal: This needs a little fix, I believe here you need to build the path string since [`os.path.exists()` only takes one argument](https://docs.python.org/3.6/library/os.path.html#os.path.exists) maybe something like: `os.path.exists(os.path.join(destination_path,f))`

tests/unit/test_datasets.py
```diff
@@ -74,12 +74,14 @@ def test_fetch(self, datasets):
...
-        datasets.return_value.downloader.download.assert_called_once_with('file.xz')
-
-    @patch('serenata_toolbox.datasets.Datasets')
     def test_fetch_latest_backup(self, datasets):
         fetch_latest_backup('test')
         self.assertTrue(datasets.return_value.downloader.download.called)
+
+    @patch('os.path.exists')
+    @patch('serenata_toolbox.datasets.Datasets')
+    def test_fetch_latest_backup_only_when_missing(self, datasets, os_path_exists):
```
jtemporal: This test should also be added to the Journey tests don't you think?

cc @cuducos @lipemorais

jtemporal:
Hi @tuliocasagrande thanks for this! Once this is done it will be of great value so, thank you!

How I tested this:

Since this PR will benefit mostly other repositories setup I used `serenata-de-amor` environment to test this:
- Edited the `requirements.txt` in serenata-de-amor repo so it would install the toolbox from @tuliocasagrande fork by replacing `serenata-toolbox` with `git+https://github.com/tuliocasagrande/serenata-toolbox.git@datasets-download`
- with `serenata_de_amor` env activated ran the `setup` file which gave me the following error:
screen shot 2017-08-25 at 8 25 19 pm
- undid the changes to the `requirements.txt` file 
- ran `setup` file again and everything works

Already commented on the review above but the problem seem only to be that `os.path.exists` needs the receive the path to the file already "built".

jtemporal:
On a general note though, and this is a 💅  more a _nice to have_ thinking on user feedback:

If none dataset needs to be downloaded a message like "You have all the latest datasets! Nothing to download." Would be awesome ;)

tests/unit/test_datasets.py
cuducos: No, I think the proper mocks @tuliocasagrande added make it a typical unit test.

serenata_toolbox/datasets/__init__.py
tuliocasagrande: Thank you @jtemporal. I introduced this error when moving to list comprehension and forgot to test manually after that.

PR #134, opened 2017-08-19
Build stages on travis

lipemorais:
The purpose of this is introduce Build Stages feature on the project to address #129 

I edit the `.travis.yml` file to use this new feature.

Here we have a [green build](https://travis-ci.org/lipemorais/serenata-toolbox/builds/265673295) using this feature.

@Irio @cuducos @jtemporal @anaschwendler could help with this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jobs:
       script: coverage run -m unittest discover tests/unit unit
       script: coverage run -m unittest discover tests/journey journey
     - stage: deploy
-      script: "echo It works!"
+      script: "echo 'It works!'"
 install:
   - python setup.py develop
   - pip install coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ cache:
     - $HOME/virtualenv/python3.5.3/
 jobs:
   include:
-    - stage: test
+    - stage: unit
       script: coverage run -m unittest discover tests/unit
+    - stage: journey
       script: coverage run -m unittest discover tests/journey
     - stage: deploy
       script: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
     - $HOME/virtualenv/python3.5.3/
 jobs:
   include:
+    - stage: test
       script: coverage run -m unittest discover tests/unit
       script: coverage run -m unittest discover tests/journey
     - stage: deploy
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,23 +1,29 @@
...
+jobs:
+  include:
+    - stage: unit
+      script: coverage run -m unittest discover tests/unit
+
+    - stage: journey
+      script: coverage run -m unittest discover tests/journey
+
+    - stage: deploy
+      script: skip
```
jtemporal: is this really necessary to be explicit? I thought not adding any would automagically skip the scripting step if not stated.

lipemorais: Sure! This is not necessary anymore. I will remove it.

anaschwendler:
Tested the PR and it seems to be working, I look into the code, but didn't know what to judge, maybe @jtemporal can do that part.

To test this PR, I clonned the project and ran into terminal: `python -m unittest discover tests`

lipemorais:
@anaschwendler would nice to try the same command `coverage run -m unittest discover tests` just to be completely sure. This command will run both, unit and journey tests.

jtemporal:
>would nice to try the same command coverage run -m unittest discover tests just to be completely sure 

Running `coverage run -m unittest discover tests` only runs unit tests. Same happens with `python -m unittest discover tests`.

Trying to figure out how to run both tests with `coverage`. Will get back here once I got anything worth sharing

jtemporal:
@cuducos just figured out! Thanks btw ❤

@lipemorais we need to add a `__init__.py` to `tests/journey` :)

That will be enough to make both test suites run =)

.travis.yml
jtemporal: When you do I'll merge this =)

lipemorais: I thought that I already had answered it but this line is necessary based on [this piece of build job documentation](https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Deployments) because each step try to run a script but it's not necessary for a deploy job.

cuducos:
Should we add a version bump or a `[ci skip]` here… I'm not sure.

jtemporal:
>Should we add a version bump or a `[ci skip]` here…? I'm not sure.

Actually we a have a conflict here since #132 was merged before this one... If @lipemorais could go ahead and fix the conflicts by updating the version to `12.1.1` I'll merge this right away.

This also means that #137 will also need some conflict solving but will be much better when taking into consideration releases.

PR #139, opened 2017-08-29
Avoids variable with one letter on __init__ for datasets

lipemorais:


cuducos:
IMHO this is not necessary. Three reasons for that:

* As this is a variable scoped into a generator, very specific and with short life it might helps readers to keep the focus in the generator's logic, not in the naming given to the interation
* [PEP8 only restricts single character names when the single character can be misleading](https://www.python.org/dev/peps/pep-0008/#names-to-avoid) (is `l` a number one, a small case L or an upper case i?)
* [PyLint actually allows single character within generators](http://pylint-messages.wikidot.com/messages:c0103)

jtemporal:
I agree with @cuducos on this one. And would like to add one other point:
- generators are a well documented structure in python so even if we don't use _significant_ naming for the variables it is acceptable

@lipemorais what do you think?

lipemorais:
I understand your points @cuducos and @jtemporal and agree that the linters and code language style don't tell us to avoid this kind of variables.

As Zen of Python says:

```
Beautiful is better than ugly.
Explicit is better than implicit.
Simple is better than complex.
...
Readability counts.
```
I see this points above as helper to keep code easy maintain, because it's what we on daily basis.

I just opened it for maintainability principles. We write once but read it a lot of times. A meaningful name makes easier this task that we do all the time, **read code**.

cuducos:
Closed in favor of #143

Issue #77, opened 2017-05-25
Become a project 100% Healthy 


lipemorais:
The purpose of this PR is have 100% of Health, fixing all coding quality analyze issues. 
[landscape build](https://landscape.io/github/datasciencebr/serenata-toolbox/147)

othman853:
I just saw that Landscape is in a hiatus for Open Source projects, meaning that they may not work for a while: https://landscape.io/hiatus

cuducos:
~meaning that they may not work for a long while~

Nope. Meaning that CI is disabled for a while. We have to trigger it manually instead.

lipemorais:
I ran prospector using the configuration file below:
```
// .prospector.yaml
output-format: grouped

strictness: high
test-warnings: true
doc-warnings: false
member-warnings: true

pep8:
  full: true
  options:
    max-line-length: 250

mccabe:
  run: true
```

There is any additional configuration that would be good?
Should it be running on CI too? I think that will be good to keep the healthy of the project 100%.

cuducos:
IMHO: 

1. As suggested in #81 strictness should be `veryhigh`, not `high`
2. As proposed in #9 `doc-warnings` should be `true` 
3. I see no reason to `max-line-length: 250`, I'd go with PEP80 default — it's better for go with something consolidates than to assume an arbitrary valu

PR #144, opened 2017-09-12
Fixed reimbursement with net value with comma

viniciusartur:
Fixing bug on Rosie reported at issue [#77](https://github.com/datasciencebr/rosie/issues/77)

Bug:
Given a CSV with a reimbursement with net_value containing comma (ex: 77,99)
When I translate this CSV using Serenata-Toolbox
Then I have a CSV translated with net_value containing comma

Expected:
Then I have a CSV translated with net_value containing dot

Bug analysis:
Converter responsible for replacing comma with dot was removed on this [commit](https://github.com/datasciencebr/serenata-toolbox/commit/9410cc16d4d70531ad0bd4990cbf82e7fefea1b8#commitcomment-23955254).

Solution:
Using decimal attribute on pandas.read_csv(), as suggested by @cuducos [here](https://github.com/datasciencebr/serenata-toolbox/commit/9410cc16d4d70531ad0bd4990cbf82e7fefea1b8#commitcomment-23955254).
Test added to cover this specific scenario.


tests/fixtures/chamber_of_deputies/reimbursements-with-comma
```diff
@@ -0,0 +1,2 @@
+congressperson_name,congressperson_id,congressperson_document,term,state,party,term_id,subquota_number,subquota_description,subquota_group_id,subquota_group_description,supplier,cnpj_cpf,document_number,document_type,issue_date,document_value,remark_value,net_value,month,year,installment,passenger,leg_of_the_trip,batch_number,reimbursement_number,reimbursement_value,applicant_id,document_id
+ABELARDO CAMARINHA,141463,329,2011,SP,PSB,54,1,Maintenance of office supporting parliamentary activity,0,,TIM CELULAR S/A,04206050005140,20272-AB,0,2009-05-19 00:00:00,411.51,0,411.51,5,2009,0,,,406387,2950,0,1772,1615640
```
rennerocha: Is this file correct? The name says that the reimbursements has a comma, but the content has a value with a point.

viniciusartur: I renamed to reimbursements-with-decimal-point. Thank you for pointing out this.

tests/fixtures/chamber_of_deputies/reimbursements-with-decimal-point
```diff
@@ -0,0 +1,2 @@
+congressperson_name,congressperson_id,congressperson_document,term,state,party,term_id,subquota_number,subquota_description,subquota_group_id,subquota_group_description,supplier,cnpj_cpf,document_number,document_type,issue_date,document_value,remark_value,net_value,month,year,installment,passenger,leg_of_the_trip,batch_number,reimbursement_number,reimbursement_value,applicant_id,document_id
+ABELARDO CAMARINHA,141463,329,2011,SP,PSB,54,1,Maintenance of office supporting parliamentary activity,0,,TIM CELULAR S/A,04206050005140,20272-AB,0,2009-05-19 00:00:00,411.51,0,411.51,5,2009,0,,,406387,2950,0,1772,1615640
```
cuducos: I guess we should have `.csv` as a extension in this filename, right? Also can we have equivalent file names?

For example, one is `…with-comma` and the other `…with-decimal-point`. What about:

* `Ano-with-comma.csv`
*  `reimbursements-with-point.csv`

Or

* `Ano-with-comma-as-decimal-separator.csv`
*  `reimbursements-with-point-as-decimal-separator.csv`

tests/unit/chambers_of_deputies/test_chamber_of_deputies_dataset.py
```diff
@@ -67,6 +68,15 @@ def test_clean_2017_reimbursements(self):
             with self.subTest():
                 assert(subquota in all_subquotas)
 
+    def test_translate_csv_with_reimbursement_with_net_value_with_comma(self):
+        csv_with_comma = os.path.join(self.fixtures_path, 'Ano-with-comma.csv')
+        with open(os.path.join(self.fixtures_path, 'reimbursements-with-decimal-point'), 'r') as csv_expected:
```
cuducos: Can we define the path in another line so we keep under the PEP8 suggestion of 80 cols?

tests/unit/chambers_of_deputies/test_chamber_of_deputies_dataset.py
```diff
@@ -67,6 +68,15 @@ def test_clean_2017_reimbursements(self):
...
+            expected = csv_expected.read()
+        
+        xz_output = Dataset('')._translate_file(csv_with_comma)
+        output = lzma.open(xz_output).read().decode('utf-8')
```
cuducos: We need a context manager here too (`with lzma.open(…)`)… or at least to close the lzma file handler after reading it.

tests/fixtures/chamber_of_deputies/reimbursements-with-decimal-point
viniciusartur: I agree with the first option. I'm working on it.

tests/unit/chambers_of_deputies/test_chamber_of_deputies_dataset.py
viniciusartur: I agree with that. I'm working on it.

tests/unit/chambers_of_deputies/test_chamber_of_deputies_dataset.py
viniciusartur: I agree with that. I'm working on it.

viniciusartur:
To ensure [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html), we changed from comma to decimal-comma.

tests/unit/chambers_of_deputies/test_chamber_of_deputies_dataset.py
```diff
@@ -67,6 +68,18 @@ def test_clean_2017_reimbursements(self):
...
+        csv_with_decimal_comma = os.path.join(self.fixtures_path, 'Ano-with-decimal-comma.csv')
+        path_with_decimal_point = os.path.join(self.fixtures_path, 
+             'reimbursements-with-decimal-point.csv')
+        with open(path_with_decimal_point, 'r') as csv_expected:
+            expected = csv_expected.read()
+        
+        xz_path = Dataset('')._translate_file(csv_with_decimal_comma)
+        with lzma.open(xz_path) as xz_file:
+            output = xz_file.read().decode('utf-8')
+        assert(output == expected)
```
cuducos: Let's use `unittest` to get better messages when test fails? That would be `self.assertEqual(output, expected)`

Issue #146, opened 2017-09-22
What should our tests do?


lipemorais:
**What is the problem?**
I can see through some PR's decisions that the definitions of tests are differents. Specially about what should journey tests cover and do.

**How can this be addressed?**
I believe that after a discussion a simple Markdown file inside the project could keep this definition inside the project.

**Who could help with this issue?**
I think that @Irio @jtemporal @cuducos @anaschwendler could contribute to this discussion.

**What I want from you?**
Share with us here answers for 3 questions:
- What should tests do and covers in general?
- What should `unit-tests` do and cover?
- What should `journey-tests` do and cover?

All of it focused on Serenata Toolbox context.

**Labels**
[question]

cuducos:
Taking my first guesses:

> What should tests do and covers in general?

The toolbox tests should assert that given an expected input (e.g. a fixture simulating the structure of a real world a dataset) the output is what we wanted.

> What should `unit-tests` do and cover?

Unit tests should test each function/method, isolating it from any environment or side-effect (use and abuse of mocks, for example).

> What should `journey-tests` do and cover?

They should depart for a controlled input (fixtures) to test the integration os all the pieces: given the input it should really create the output file, for example — without depending on any external eventuality (mock the input if it comes from a remote server, for example).

lipemorais:
> What should tests do and covers in general?

The Serenata Toolbox tests should protect us from our on mistakes providing safety to move on and help us keep everything working while we keep improving our project.

> What should unit-tests do and cover?

The unit-tests should cover the smallest unit of code. Unless it covers a piece of code that don't have coverage. What happens on project that do not have 100% of code coverage and is usual.

> What should journey-tests do and cover?

In my vision it has two principal values:
- Test real journey based on real value **but mocked**
- Test real journey based on **real integration** before we deploy to assure that everything is ok with our integration.

cuducos:
> Test real journey based on **real integration** before we deploy to assure that everything is ok with our integrations.

Experience has shown us that relying on external services (specially on government one I might add) is problematic: they are unstable, they change format/encoding without warning, etc.

This argument actually  goes both ways: on the one hand we must get used to see our CI red not because we messes up, but because these external servers did it. On the other hand we must be sure the toolbox works with them before releasing a new version.

What really bugs me is that the instability is way more common than unwanted changes in their side… so we'll be blocked most of the times just because of low availability of third-parties.

lipemorais:
[...]

PR #147, opened 2017-09-23
Improve project health

giovanisleite:
Improve the project health

Code smells pointed out by prospector was removed/fixed as like some styles that wasn't right according to pep8 style guide

Running `prospector -s veryhigh --max-line-length 120  serenata_toolbox`, it was pointed out 60 problems, now it is showing 30 (details about later).
EDIT: now 22, thanks to @lipemorais and @cuducos 

@lipemorais and anybody who wants it =))

**TODO**
- [x] Remove all code smells showed by landscape
- [ ] Remove all code styles pointed by prospector

why i didn't solve the last 30 remaining:
13 invalid-name "df" - I think the readability of "df" is just fine, (i don't have exp) but seems that it is a common practice (maybe change to what_this_df_is_about_df?)
1 invalid-name "s3" - it is literal I think.
[1 invalid-name "fh"](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/datasets/downloader.py#L109) - I couldn't understand what it means, so I couldn't think a valid name for it.
The other messages was about "too few public methods" and "no self use, could be a function" probably smells for great changes, related to Issue #87.

Landscape pointed out two more things that prospector didn't, list comprehension over filter and map functions, I changed just a map function for readability (I think), list comprehension over filters make it verbose and not more readable (if it seems inconsistency, let me know).

Should I change bump version?


lipemorais:
> 1 invalid-name "fh" - I couldn't understand what it means, so I couldn't think a valid name for it.

I believe that @cuducos may help us with it. 
I believe that it should be just f since it is the file read by `aiofiles.read`.

> Should I change bump version?

Yeah we are trying to have a at least a minor bump version for any change in this project. :)

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -27,12 +27,16 @@ def fetch(self):
...
+                    file_path,
+                    http_error_exception))
                 raise http_error_exception
             except URLError as url_error_exception:
                 print("The server couldn\'t fulfill the request.")
                 print('Reason: ', url_error_exception.reason)
-                print("While fetching, Seranata Toolbox didn't find file: {} \n{}".format(file_path, url_error_exception))
+                print("While fetching, Seranata Toolbox didn't find file: {} \n{}".format(
+                    file_path,
+                    url_error_exception))
```
cuducos: It feels like you missed one space, and we have a matter os style: 

Either… 

```python
print("While fetching, Seranata Toolbox didn't find file: {} \n{}".format(file_path,
                                                                          url_error_exception))
```

…or…

```python
print("While fetching, Seranata Toolbox didn't find file: {} \n{}".format(
    file_path,
    url_error_exception)
)
```

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -49,7 +53,9 @@ def translate(self):
             try:
                 self._translate_file(csv_path)
             except FileNotFoundError as file_not_found_error:
-                print("While translating, Seranata Toolbox didn't find file: {} \n{}".format(csv_path, file_not_found_error))
+                print("While translating, Seranata Toolbox didn't find file: {} \n{}".format(
+                    csv_path,
+                    file_not_found_error))
```
cuducos: It feels like you missed one space, and we have a matter os style: 

Either… 

```python
print("While translating, Seranata Toolbox didn't find file: {} \n{}".format(file_path,
                                                                             file_not_found_error))
```

…or…

```python
print("While translating, Seranata Toolbox didn't find file: {} \n{}".format(
    file_path,
    file_not_found_error)
)
```

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -121,10 +127,14 @@ def _translate_file(self, csv_path):
...
-            'Contratação de consultorias, assessorias, pesquisas, trabalhos técnicos e outros serviços de apoio ao exercício do mandato parlamentar':
-                'Recruitment of consultancies, advisory services, research, technical work and other services in support of the exercise of the parliamentary mandate',
+            ('Aquisição de material de consumo para uso no escritório político, inclusive aquisição ou '
+            'locação de software, despesas postais, aquisição de publicações, locação de móveis e de equipamentos. '):
+                ('Acquisition of consumables for use in the political office, including acquisition or '
+                'leasing of software, postal expenses, acquisition of publications, rental of furniture and equipment'),
+            ('Contratação de consultorias, assessorias, pesquisas, trabalhos técnicos e outros serviços'
+            ' de apoio ao exercício do mandato parlamentar'):
+                ('Recruitment of consultancies, advisory services, research, technical work and other services'
+                ' in support of the exercise of the parliamentary mandate'),
```
cuducos: I think a better option would be to define two tuples, one with the pt-BR version, another with the en-US version and finally build the dictionary with something like `categories = dict(zip(pt_br, en_us))`.

cuducos:
Ow… another doubt:

>  Remove all code smells showed by landscape
>  Remove all code styles pointed by prospector

`prospector` is the engine running in the backend of Landscape.io, so there should be no difference between these reports. If is there probably we have something missing in our [`.landscape.yml`](https://github.com/datasciencebr/serenata-toolbox/blob/master/.landscape.yml).

giovanisleite:
> When did you get the too few public methods? Probably they can be fixed with the @staticmethod decorator. Have you tried? — IMHO this bit doesn't have to do with #87, but the other warning has ; )
pylint complains about classes having less than two public methods and the datasets classes have just fetch as public.
The other complaim, about the no self use, was resolved with your tip, thanks.

> prospector is the engine running in the backend of Landscape.io, so there should be no difference between these reports. If is there probably we have something missing in our .landscape.yml.
In fact, on Landspcape.io appeared some "use list comprehension over map and filter" that prospector doesn't show up.

tests/journey/test_chamber_of_deputies_official_missions_dataset.py
```diff
@@ -20,7 +20,6 @@ def test_fetch(self):
             'canceled', 'report_status', 'report_details_link'
         ]
         self.assertTrue(np.array_equal(expectedColumns, actualColumns))
-        self.assertEqual(56, len(df))
```
cuducos: I totally agree that this test is a problem _per se_. However just for maintainability sake and for the community to follow the work on the project I'm not in favor of “smuggling” this change in the middle of a PR related to something else (code quality, health etc.).

There is a greater discussion about the [role of tests in the project](https://github.com/datasciencebr/serenata-toolbox/issues/146) going on.

Would you mind reverting this change and let this to another PR/discussion?

BTW your commits are larger than I would recommend: not in terms of amount of changes, but in the purpose of the changes committed:

* ace718d: identified static methods and removed test that doesn't make sense anymore
* d2c4622: fixed spaces according to pep8 e128 and updated minor patch


In the first one listed you committed this change together with another unrelated  change, so you cannot use `git revert` or `git rebase` to fix that. In the second is some other PR gets merged first, you cannot use `git` to revert a version bump… In general my advice is: if your commit message is _X and Y_ you should commit twice: first commit _X_ and then commit _Y_ ; )

lipemorais: I do agree. :)

giovanisleite: I removed the test with the thought that I was removing unused code as part of the project health. But I agree, it was bad and it is not consistent with others changes.

Oh, that's true. The commits are terrible.

I will fix that and come back
Thank you and srry for the trouble, like I told @lipemorais, the project that is contributing to me hahaha

Issue #148, opened 2017-10-02
Add PyPI label to README


jtemporal:
**What is the problem?**
We currently do not have our pipy version badge on the README, something like the image bellow:

<img width="105" alt="screen shot 2017-10-02 at 12 04 53 pm" src="https://user-images.githubusercontent.com/6595551/31084706-881e1462-a76b-11e7-98a5-8bd65de268c4.png">

**How can this be addressed?**
A pull request adding the pipy badge to README.md maybe using [Shields](https://shields.io)

**Who could help with this issue?**
@anaschwendler @cuducos @jtemporal

**Labels**
[hacktoberfest] [first-timers-only]

marius92mc:
I will try this. 👍

PR #149, opened 2017-10-02
Add PyPI label to README.rst

marius92mc:
Fixes https://github.com/datasciencebr/serenata-toolbox/issues/148. 

Add a URL to the mentioned label. 

See README.rst.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,6 +14,9 @@
    :target: https://coveralls.io/github/datasciencebr/serenata-toolbox?branch=master
    :alt: Coveralls
 
+.. image:: https://badge.fury.io/py/serenata-toolbox.svg
+   :alt: PyPI package version
+
 Serenata de Amor Toolbox
 ========================
 
--- a/setup.py
+++ b/setup.py
@@ -34,5 +34,5 @@ setup(
         'serenata_toolbox.datasets'
     ],
     url=REPO_URL,
-    version='12.1.7'
+    version='12.1.8'
 )


marius92mc:
Should this be merged in this state knowing the coverage remained at that rate, @giovanisleite?

giovanisleite:
> Should this be merged in this state knowing the coverage remained at that rate, @giovanisleite?

Yes, the coverage and the travis error are not related to this issue.

I don't know if it is necessary to bump  the version, but @jtemporal can tell you.

marius92mc:
Yeah, I also think it is not needed to bump the version.

giovanisleite:
> Yeah we are trying to have a at least a minor bump version for any change in this project. :)

Actually, the readme/documentation is part of the project, so.. i think it is

marius92mc:
Alright, @giovanisleite, I bumped the version.

cuducos:
Regarding the version bump issue, it's written in the `README.md`:

> Always suggest a version bump.

Always. Because every new code merged into `master` triggers the CI. And then the CI [triggers a new release to PyPI](https://github.com/datasciencebr/serenata-toolbox/blob/master/.travis.yml#L15-L23). And this would ruin the whole project (omg I'm so dramatic…) hahaha… anyway, no, that's not a big big problem. But it's important to understand that not bumping the version will break something, and it's also important to understand why it breaks something.

So unless you [add `[skip ci]` to your commit message](https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build), yes, bumping version is required ; )

Issue #42, opened 2017-04-26
Add timeout or similar to Downloader


jtemporal:
In case the download of one or more dataset is stuck we need to have a timeout or some other sort of breaking the wait gracefully.

arioston:
Hi,

What you need is a simple timeout and an error message or need something more elaborate ?  

Remove ["timeout=None"](http://aiohttp.readthedocs.io/en/stable/client.html#timeouts) will make the connection use default timeout, this is one solution.

cuducos:
I think the best approach wouldn't be to simply _remove_ [this `timeout=None`](https://github.com/datasciencebr/serenata-toolbox/blob/master/serenata_toolbox/datasets/downloader.py#L107). Instead I'd say that we user instantiate `Dataset` they could use a kwarg to customize it if they want. If they don't pass we assume it to be `None`. What do you think?

cuducos:
Closed by #160

Issue #39, opened 2017-04-12
Test async part of Datasets module


cuducos:
**If #38 is merged** the coverage points to a non-test part that is basically the coroutines:

```
Name                                          Stmts   Miss  Cover
-----------------------------------------------------------------
serenata_toolbox/datasets/__init__.py            26      0   100%
serenata_toolbox/datasets/contextmanager.py       5      0   100%
serenata_toolbox/datasets/downloader.py          58     25    57%
serenata_toolbox/datasets/local.py               15      0   100%
serenata_toolbox/datasets/remote.py              21      0   100%
-----------------------------------------------------------------
TOTAL                                           125     25    80%
```

I opened the PR as it is because I've never written unit tests for `asyncio` and studying that was postponing the new module further than I could handle. Sorry about that, guys.

PR #157, opened 2017-10-27
#42 timeout

arioston:
The timeout was added to Dataset and pass down to Download.Tests have been changed to add the new parameter, and a new test has been added to validate the timeout.

### DIFF
--- a/serenata_toolbox/datasets/__init__.py
+++ b/serenata_toolbox/datasets/__init__.py
@@ -41,7 +41,8 @@ class Datasets:
     local datasets that are not present in the remote bucket.
 
     :param local_directory: (str) path to local directory of the datasets
-    :param timeout: (float) timeout parameter to ClientSession.get(), None or 0 disables timeout check.
+    :param timeout: (float) timeout parameter to Downloader,
+           None or 0 disables timeout check.
     """
 
     def __init__(self, local_directory=None, timeout=None):
@@ -53,8 +54,8 @@ class Datasets:
         self.downloader = Downloader(
             local_directory,
             bucket=self.remote.bucket,
-            **self.remote.credentials,
-            timeout=timeout
+            timeout=timeout,
+            **self.remote.credentials
         )
 
     @property
--- a/serenata_toolbox/datasets/__init__.py
+++ b/serenata_toolbox/datasets/__init__.py
@@ -41,9 +41,10 @@ class Datasets:
     local datasets that are not present in the remote bucket.
 
     :param local_directory: (str) path to local directory of the datasets
+    :param timeout: (float) timeout parameter to ClientSession.get(), None or 0 disables timeout check.
     """
 
-    def __init__(self, local_directory=None):
+    def __init__(self, local_directory=None, timeout=None):
         if not local_directory:
             local_directory = 'data'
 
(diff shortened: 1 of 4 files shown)


serenata_toolbox/datasets/__init__.py
```diff
@@ -41,9 +41,10 @@ class Datasets:
     local datasets that are not present in the remote bucket.
 
     :param local_directory: (str) path to local directory of the datasets
+    :param timeout: (float) timeout parameter to ClientSession.get(), None or 0 disables timeout check.
```
cuducos: Would you mind keeping lines under 80 chars as PEP8 suggests?

serenata_toolbox/datasets/__init__.py
```diff
@@ -52,7 +53,8 @@ def __init__(self, local_directory=None):
         self.downloader = Downloader(
             local_directory,
             bucket=self.remote.bucket,
-            **self.remote.credentials
+            **self.remote.credentials,
+            timeout=timeout
```
cuducos: Would you mind passing all specified keyword arguments before unpacking the others? That is to say:

```python
Downloader(local_directory, bucket=self.remote.bucket, timeout=None,  **self.remote.credentials)
```

instead of

```python
Downloader(local_directory, bucket=self.remote.bucket, **self.remote.credentials, timeout=None)
```

tests/unit/test_datasets_downloader.py
```diff
@@ -1,10 +1,13 @@
-import asyncio
```
cuducos: Would you mind keeping [imports ordered as suggested in PEP8](https://www.python.org/dev/peps/pep-0008/#imports)? First native modules, then third-party and so on… you can use tools as `isort` or `prospector` to help you out ; )

tests/unit/test_datasets_downloader.py
```diff
@@ -84,7 +94,7 @@ def test_download_single_file(self, exists, isdir, asyncio_, main):
         loop = asyncio_.get_event_loop.return_value
         self.assertTrue(loop.run_until_complete.called)
         main.assert_called_once_with(loop, ('test.xz',))
-
+
```
cuducos: Would you mind removing extra blank spaces in this supposed to be empty line?

tests/unit/test_datasets_downloader.py
```diff
@@ -1,10 +1,13 @@
...
 import os
 from unittest import TestCase
 from unittest.mock import Mock, patch
 
+import asyncio
+
+import aiohttp
 from aiohttp.test_utils import make_mocked_request
 from aiohttp.web import Response
+from aiohttp.client_exceptions import TimeoutError
```
cuducos: Maybe we have too many imports from `aiohttp`. Either you keep what the module was originaly using (e.g. `from aiohttp import ClientSession` to access `ClientSession` – instead of `import aiohttp` and then `aiohttp.ClientSession`) **or** you remove extra imports and access everything from the main `aiohttp` module.

arioston:
Hi @cuducos, I think this PR only made a test of the timeout, but after finish this, I will start on #39 and research a better way to test the other scenarios.

cuducos:
Sounds great, @arioston! Now all that's missing is [version bump](https://github.com/datasciencebr/serenata-toolbox#contributing) ; )

PR #162, opened 2017-11-29
Fix asyncio error in tests

cuducos:
**What is the purpose of this Pull Request?**
Last build with the updated version of `aiohttp` has [broken our tests CI](https://travis-ci.org/datasciencebr/serenata-toolbox/builds/309092425). 

**What was done to achieve this purpose?**
This PR adapts our code to the new error architecture of current version of `aiohttp`, using `TimeoutError` from `concurrent.futures` instead of `aiohttp.client_exceptions`.

**How to test if it really works?**
`$ pytest` or try to download something and hope the server takes too long to cause a timeout ; )

**Who can help reviewing it?**
@anaschwendler

PR #156, opened 2017-10-15
Replaces prints with logging functions

jcemelanda:
This pullrequest replaces the print calls throughout the code with the logging module.

I believe this review could be done by anyone with deep enough understanding of the code to know if the logging methods called are really the best suited ones for each place.

I tried to define the log level of each call based on the code around each print call. Usually I used logging.info, errors when dealing with errors and exceptions and warnings when I saw the message was a warning message.

Fixes #25

PR #163, opened 2017-11-29
Use logs instead of print (extends #156)

cuducos:
This PR is based on #162 in order to be able to run tests without error. **So it's recommended to review #162 before this one**.

---

This is a follow up for #156  — after code review I'm afraid @jcemelanda couldn't get back to the code, so we jumped in because we wanted these enhancements so bad ; ) In addition we added a custom logger to be used anywhere and this log is automatically outputting messages to `stdout`.

* When using `log.debug` we deleted the `if os.environ.get('DEBUG')` because it was already explicit it's a debug message/log
* We enhanced some error/info messages with failing URL and useful information about the event being registered in the log
* Added and setup a logger a `serenata_toolbox/__init__.py` to be imported and used within other modules

Tests cannot fail, plus running some scripts (the 11 changed) and checking logs are there, for example:

pasted_image_29_11_17_18_53

@anaschwendler

### DIFF
--- a/serenata_toolbox/__init__.py
+++ b/serenata_toolbox/__init__.py
@@ -1,0 +1,18 @@
+import logging
+import os
+import sys
+
+
+LEVEL = logging.DEBUG if os.environ.get('DEBUG') else logging.INFO
+LOG_FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+
+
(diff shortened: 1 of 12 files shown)


anaschwendler:
Hello @cuducos just did it, and it is working! 🎉

PR #179, opened 2018-02-20
Refactor the way Chamber of Deputies' reimbursements are downloaded and cleaned

Irio:
For quite some time, working on these classes have been taking more effort than expected.
This refactoring comes to address multiple issues: code complexity, memory usage and
storing, multiple times, the same dataset in disk.

How to use it:

```python
from serenata_toolbox.chamber_of_deputies.reimbursements import \
	Reimbursements

down = Reimbursements(2018)
print('csv path', down())
```

**TODO**
- [x] Investigate the test failing in the CI but passing in my env.


serenata_toolbox/chamber_of_deputies/reimbursements.py
```diff
@@ -1,133 +1,39 @@
...
+
+    def __init__(self, year, path=None):
+        self.year = year
+        self.path = path or gettempdir()
+
+    def __call__(self):
+        self.fetch()
+        self.clean()
+        file_path = os.path.join(
+            self.path, 'reimbursements-{}.csv'.format(self.year))
```
cuducos: As we're using Python 3.6 we could make it a bit more readable using the new f-strings:

```python
file_path = os.path.join(self.path, f'reimbursements-{self.year}.csv')
```

serenata_toolbox/chamber_of_deputies/reimbursements.py
```diff
@@ -1,133 +1,39 @@
...
+        return file_path
+
+    def fetch(self):
+        file_path = os.path.join(self.path, 'Ano-{}.zip'.format(self.year))
```
cuducos: As we're using Python 3.6 we could make it a bit more readable using the new f-strings:

```python
file_path = os.path.join(self.path, 'Ano-{self.year}.zip')
```

serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py
```diff
@@ -0,0 +1,173 @@
...
+        self.data = None
+
+    def __call__(self):
+        self.load_source_file()
+        self.translate()
+        self.aggregate_multiple_payments()
+        self.save()
+
+    def load_source_file(self):
+        file_path = os.path.join(self.path, 'Ano-{}.csv'.format(self.year))
```
cuducos: As we're using Python 3.6 we could make it a bit more readable using the new f-strings:

```python
file_path = os.path.join(self.path, f'Ano-{self.year}.csv')
```

serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def aggregate_multiple_payments(self):
+        self.data = pd.concat([
+            self._house_payments(),
+            self._non_house_payments(),
+        ])
+
+    def save(self):
+        file_path = os.path.join(
+            self.path, 'reimbursements-{}.csv'.format(self.year))
```
cuducos: Two repeated comments: 

**1**

As we're using Python 3.6 we could make it a bit more readable using the new f-strings:

```python
file_path = os.path.join(self.path, f'reimbursements-{self.year}.csv')
```
**2**

Maybe here is a nice place to use `print(file_path)` to inform users where the dataset was actually saved (having in mind that it could be a random tmp dir).

setup.py
```diff
@@ -35,5 +35,5 @@
         'serenata_toolbox.datasets'
     ],
     url=REPO_URL,
-    version='12.4.4'
+    version='13.0.0'
```
cuducos: Not related to the version but, but to `setup.py`: If we had to update tests from Python 3.5 to Python 3.6, we might need to change line 13 to remove support for Python 3.5 and explicitly support for Python 3.6.

cuducos:
>  Investigate the test failing in the CI but passing in my env.

`TypeError: Use async with instead` seams to be from `aiohttp`, the async library to handle async HTTP requests in the toolbox. The [`__enter__` method raises that error pointnig us to use `__aenter__`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L113-L125). In other words, pushing us to use the async version of the context manager:

```pyhton

# this
async with foobar():
    pass

# instead of
with foobar():
    pass
```

**TLDR** Probably we need to add `async` to [this `with` block](https://github.com/okfn-brasil/serenata-toolbox/blob/master/tests/unit/test_datasets_downloader.py#L145-L146):

```python
            async with ClientSession(loop=loop) as client:
                yield from downloader.fetch_file(client, '2016-12-06-reibursements.xz')
```

serenata_toolbox/chamber_of_deputies/reimbursements.py
```diff
@@ -1,133 +1,39 @@
...
-        df.to_csv(filepath, **self.CSV_PARAMS)
-
-        log.info('Done.')
+    """
+    Get an updated version of the reimbursements dataset for a given year.
+    """
+
+    def __init__(self, year, path=None):
+        self.year = year
+        self.path = path or gettempdir()
```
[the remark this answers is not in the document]
Irio: `.__call__` returns the path of the CSV.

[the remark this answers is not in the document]
cuducos: True. My bad!

Irio:
Unit tests passing but timeout on the journey one.

cuducos:
> Unit tests passing but timeout on the journey one.

Running it locally just in case. I would just push the `travis_wait` further (something absurd just to check if it's a real timeout or an error; then fine tune it to something reasonable, maybe using `time` if it's available in the CI).

cuducos:
Passed locally here too, probably a higher number in the `travis_wait` is gonna fix the broken CI.

PR #194, opened 2018-08-29
Update README.rst

willianpaixao:
* Rosie link updated.
* Better explanation on creating virtual environments.
* Python commands replaced by Python3.

__To have a more detailed and updated documentation for contribution.__

__Change the README.rst file.__

__Copy the commands and run locally.__

__Downloading the code and building it locally, following the tutorial.__

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,14 +25,14 @@ Serenata de Amor Toolbox
 ========================
 
 `pip <https://pip.pypa.io/en/stable/>`_  installable package to support `Serenata de Amor <https://github.com/okfn-brasil/serenata-de-amor>`_
-and `Rosie <https://github.com/okfn-brasil/rosie>`_ development.
+and `Rosie <https://github.com/okfn-brasil/serenata-de-amor/blob/master/rosie/README.md>`_ development.
 
 Serenata_toolbox is compatible with Python 3.6+
 
 Installation
 ------------
 
-::
+.. code-block:: bash
 
     $ pip install -U serenata-toolbox
 
@@ -48,7 +48,7 @@ Example 1: How do I download the datasets?
 
 We have `plenty of them <https://github.com/okfn-brasil/serenata-de-amor/blob/master/research/CONTRIBUTING.md#datasets-data>`_ ready for you to download from our servers. And this toolbox helps you get them. Let's say you want your datasets at `data/`:
 
-.. code:: python
+.. code-block:: python
 
   from serenata_toolbox.datasets import Datasets
   datasets = Datasets('data/')
@@ -69,7 +69,7 @@ Example 2: Using shortcuts
 
 If the last example doesn't look that simple, there are some fancy shortcuts available:
 
-.. code:: python
+.. code-block:: python
 
   from serenata_toolbox.datasets import fetch, fetch_latest_backup
   fetch('2018-01-05-reimbursements.xz', 'data/')
@@ -80,7 +80,7 @@ Example 3: Generating datasets
 
 If you ever wonder how did we generated these datasets, this toolbox can help you too (at least with the more used ones — the other ones are generated `in our main repo <https://github.com/okfn-brasil/serenata-de-amor/blob/master/research/CONTRIBUTING.md#the-toolbox-and-our-the-source-files-src>`_):
 
-.. code:: python
+.. code-block:: python
 
     from serenata_toolbox.federal_senate.dataset import Dataset as SenateDataset
     from serenata_toolbox.chamber_of_deputies.reimbursements import Reimbursements as ChamberDataset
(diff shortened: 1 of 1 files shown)


willianpaixao:
I just would like to better explain the reason why I requested the change from `python`and `pip` to `python3`and `pip3`.
As Python 3 isn't the default in almost all OSes, the user must explicitly install and change the default in order to call Python 3 under the command `python`. [[1]](https://unix.stackexchange.com/questions/410579/change-the-python3-default-version-in-ubuntu)

More advanced users when facing the errors for building the code under Python 2.x would quickly understand and correct the commands but would be nice to make easier to use the more compatible command in the documentation.

README.rst
```diff
@@ -98,45 +98,48 @@ Documentation (WIP)
...
+.. code-block:: bash
 
   $ cd docs
   $ make clean;make rst;rm source/modules.rst;make html
 
 Contributing
 ------------
 
-Within your `virtualenv <https://virtualenv.pypa.io/en/stable/>`_:
+Start by cloning the repository and installing the dependencies. You should create a development environment with `virtualenv <https://virtualenv.pypa.io/en/stable/>`_ by running:
```
cuducos: Actually as this is a `pip` installable project there's no _installing the dependencies_ step. The `python setup.py develop` takes care of that (and actually installing the dependencies otherwise can lead to issues in the development process).

Note that in the bash steps ahead there's no line to install the dependencies.

README.rst
```diff
@@ -98,45 +98,48 @@ Documentation (WIP)
...
 
-::
+.. code-block:: bash
 
   $ git clone https://github.com/okfn-brasil/serenata-toolbox.git
+  $ cd serenata-toolbox
+  $ virtualenv .env
+  $ source .env/bin/activate
```
cuducos: Usually I don't include virtualenv options because different setups might have different ways to crate virtualenvs (`virtualenv`, `python -m venv`, `pipenv`, `mkvirtualenv` from `virtualenv-wrapper`). **If** you think it's really important the default for Python 3.6 is `python -m venv` I guess (but need to be checked in the docs).

README.rst
willianpaixao: I'll change that.

README.rst
willianpaixao: IMHO I think it's important to describe _all_ the commands necessary to run the application. Create and activate the dev environment is very basic and any long-time developer would do that without hesitating, but I'm thinking about the beginners who want to contribute.

willianpaixao: What should I do? You have the final call.

cuducos: Just checked [the docs](https://docs.python.org/3/library/venv.html#creating-virtual-environments): the command would be `python -m venv .env`, not `virtualenv .env` — so I'll keep the _requested changes_ status.

But more important than this detail is what follows:

Anyway, I'm also _thinking about the beginners who want to contribute_ — if you think otherwise you're getting me completely wrong. However my point is: if they just copy and past commands without any understanding of what's going on this is not _necessarily_ the best way to bring them onboard. If our commands happen to not work for their system, this generates noise and discourage them to move on. 

I do prefer to indicate _you have to create a virtualenv_ and let them do it the way they have already learned, or do their own research according to their system and needs. So we avoid commands that might not work for some users. 

Wanna any use cases where this approach is better than `virtualenv .env` plus `source .env/bin/activate`? Here we go:  I was told calling the virtualenv `.env` might not work in some Windows version; but as a matter of fact `source .env/bin/activate` does not work on Windows at all! And [49% of Python developers use Windows](https://www.jetbrains.com/research/python-developers-survey-2017/#developer-profile) – yet, 17% use other OS than macOS, Linux and Windows.

So I do understand you're willing _to describe all the commands_ but I don't think that's feasible (or, if so, if it's desirable because our docs would be too verbose full of special cases). That's why in the docs I tend to focus on what is specific to the application I'm documenting and let the language- and environment-specific instructions for users to do their own research according to their context and needs.

cuducos: What about linking to the official Python docs on `venv`?

---

Within your [`venv`](https://docs.python.org/3/library/venv.html#creating-virtual-environments) clone the repository and install the package for development:

```sh
$ git clone https://github.com/okfn-brasil/serenata-toolbox.git
$ cd serenata-toolbox
$ python setup.py develop
```
---

What do you think?

willianpaixao:
All changes made and commits squashed.

README.rst
```diff
@@ -98,45 +98,47 @@ Documentation (WIP)
...
+.. code-block:: bash
 
   $ cd docs
   $ make clean;make rst;rm source/modules.rst;make html
 
 Contributing
 ------------
 
-Within your `virtualenv <https://virtualenv.pypa.io/en/stable/>`_:
+Firstly, you should create a development environment with `virtualenv <https://virtualenv.pypa.io/en/stable/>`_ to isolate your development.
```
cuducos: Didn't you like [my suggestion to update the reference link to Python official docs](https://github.com/okfn-brasil/serenata-toolbox/pull/194#discussion_r213675625) (as well as replacing `virtualenv` by `venv` as it's called in Python 3.6)?

willianpaixao: Updated.

PR #201, opened 2018-09-07
[WIP] Command line wrapper

willianpaixao:
This is a POW for the CLI wrapper.
Nothing fancy, it just calls the already existent classes, passing the arguments down the chain, then triggering the download.

I want to know if the choice for the executable `serenata-toolbox` and the parameters `--dataset`, `--year` (next to be implemented)  are good choices. Let me know if you have a better suggestion.


README.rst
```diff
@@ -46,7 +46,26 @@ Usage
 Example 1: How do I download the datasets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-We have `plenty of them <https://github.com/okfn-brasil/serenata-de-amor/blob/master/research/CONTRIBUTING.md#datasets-data>`_ ready for you to download from our servers. And this toolbox helps you get them. Let's say you want your datasets at `data/`:
+We have `plenty of them <https://github.com/okfn-brasil/serenata-de-amor/blob/master/research/CONTRIBUTING.md#datasets-data>`_ ready for you to download from our servers. And this toolbox helps you get them. Here some examples:
+
+.. code-block:: bash
+
+    # without any arguments will download pre-propre-processedd datasets and store into export/ folder
```
cuducos: There's a typo (`pre-propre-processedd`) but the main question is why changing the default to `export/`?

Usually we use Rosie's default (`/tmp/serenata-data`) or [the toolbox's default](https://github.com/okfn-brasil/serenata-toolbox/blob/master/serenata_toolbox/datasets/__init__.py#L33-L35) (`data/`).

Can we keep the toolbox's default instead of making the script behave one way and the CLI another way?

README.rst
```diff
@@ -46,7 +46,26 @@ Usage
...
+    $ serenata-toolbox
+
+    # will download these specific datasets and store into export/ folder
+    $ serenata-toolbox export --dataset federal-senate reimbursements
```
cuducos: Not clear what is happening here. Does `federal-senate reimbursements` stands for reimbursements from the Federal Senate, or data from the federal senate (which is called federal-senate.xz) as well as data from the chamber of deputies (which is called reimbursements-YYYY.xz)?

I would rather using the [module names](https://github.com/okfn-brasil/serenata-toolbox/tree/master/serenata_toolbox) for clarity – but this implementation raises an interesting issue on renaming the generated dataset to avoid this confusion, right? Maybe in another issue/PR ; ) 

Also, what is the role of export in this command? Is this the directory where data is to be saved, or a proper command?

serenata_toolbox/serenata-toolbox
```diff
@@ -0,0 +1,52 @@
...
+
+def main():
+    args = parser()
+    path = args.path
+    if not os.path.isdir(path):
+        try:
+            os.mkdir(os.path.join(path))
+        except OSError:
+            log.error('Couldn\'t create the output directory')
+            raise
```
cuducos: As the `Dataset` class has a default directory, we could implement this `try`/`except` block in its `__init__` and make the CLI and the script behavior similar.

serenata_toolbox/serenata-toolbox
```diff
@@ -0,0 +1,52 @@
...
+    if not args.dataset:
+        datasets = Datasets(path)
+        datasets.downloader.download(datasets.downloader.LATEST)
+        return
+    if 'reimbursements' in args.dataset:
+        if args.year:
+            chamber = ChamberDataset(path, args.year)
+        else:
+            chamber = ChamberDataset(path)
```
cuducos: I guess if the user doesn't pass `--year` it becomes `None`. So there's no problem in calling `ChamberDataset(path, args.year)` in both cases. But this is merely a guess that needs to be checked.

setup.py
```diff
@@ -36,7 +36,8 @@
         'serenata_toolbox.chamber_of_deputies',
         'serenata_toolbox.datasets'
     ],
+    scripts=['serenata_toolbox/serenata-toolbox'],
     url=REPO_URL,
     python_requires='>=3.6',
-    version='15.0.5',
+    version='15.0.6',
```
cuducos: This is a minor (not a micro) bump: _values have been added, existing values are unchanged_. 15.1.0 then.

README.rst
willianpaixao: 1 It's a typo and 2 I can change the default folder.

willianpaixao: BTW `export is just an example of changing the already default `data` folder.

README.rst
willianpaixao: Ok, changing to modules. And yes, `export` is the example for output directory.

cuducos: > And yes, `export` is the example for output directory.

Just an idea; I'd add something with slashes to be clearer (maybe Rosie's default)

```sh
$ serenata-toolbox /tmp/serenata-data --dataset federal-senate chamber-of-deputies
```

serenata_toolbox/serenata-toolbox
willianpaixao: Then you want me to remove this part here and add in `__init__` add there, having the check in both places?

serenata_toolbox/serenata-toolbox
willianpaixao: Hum, I think I got some error when I tried that. Let me try again.

serenata_toolbox/serenata-toolbox
cuducos: > having the check in both places?

No, we can just check if the directory exists only once, in `Dataset` ; )

serenata_toolbox/serenata-toolbox
cuducos: So maybe `chamber = ChamberDataset(path, args.year or None)`

serenata_toolbox/chamber_of_deputies/reimbursements.py
```diff
@@ -19,9 +18,15 @@ class Reimbursements:
...
-    def __init__(self, year, path=None):
+    def __init__(self, year='2018', path=None):
         self.year = year
-        self.path = path or gettempdir()
+        if not os.path.isdir(path):
+            try:
+                os.mkdir(os.path.join(path))
+            except OSError:
+                log.error('Couldn\'t create the output directory')
+                raise
```
cuducos: We should be explicit about which exception we're raising here. We can either throw the default `OSError` or build our custom exception.

TBH I would simply and just trust the `OSError`:

```python
if not os.path.isdir(path):
    os.mkdir(os.path.join(path))
```

With no `try`/`exception` at this point. What do you think about it?

serenata_toolbox/federal_senate/dataset.py
```diff
@@ -14,6 +14,12 @@ class Dataset:
     AVAILABLE_YEARS = [year for year in range(2008, date.today().year + 1)]
 
     def __init__(self, path, years=AVAILABLE_YEARS):
+        if not os.path.isdir(path):
+            try:
+                os.mkdir(os.path.join(path))
+            except OSError:
+                log.error('Couldn\'t create the output directory')
+                raise
```
cuducos: We should be explicit about which exception we're raising here. We can either throw the default `OSError` or build our custom exception.

TBH I would simply and just trust the `OSError`:

```python
if not os.path.isdir(path):
    os.mkdir(os.path.join(path))
```

With no `try`/`exception` at this point. What do you think about it?

serenata_toolbox/serenata-toolbox
```diff
@@ -0,0 +1,46 @@
...
+
+
+def main():
+    args = parser()
+    path = args.path
+    if not args.module:
+        datasets = Datasets(path)
+        datasets.downloader.download(datasets.downloader.LATEST)
+        return
+    if 'reimbursements' in args.module:
```
cuducos: I do believe we should use module name (`chamber_of_deputies`) instead of the dataset prefix (`reimbursements`). What do you think about it?

serenata_toolbox/serenata-toolbox
```diff
@@ -0,0 +1,46 @@
...
+        if args.year:
+            chamber = ChamberDataset(path=path, year=args.year)
+        else:
+            chamber = ChamberDataset(path=path)
+        chamber()
+    if 'federal-senate' in args.module:
```
cuducos: I do believe we should use module name (`federal_senate`) instead of the dataset prefix (`federal-senate`). What do you think about it?

README.rst
```diff
@@ -43,10 +43,32 @@ If you are a core developer willing to upload datasets to the cloud you need to
...
+    $ serenata-toolbox
+
+    # will download these specific datasets and store into /tmp/serenata-data folder
+    $ serenata-toolbox /tmp/serenata-data --module federal-senate reimbursements
+
+    # you can specify a dataset and a year
+    $ serenata-toolbox --module reimbursements --year 2009
+
+    # or specify all options simultaneously
+    $ serenata-toolbox /tmp/serenata-data --module federal-senate --year 2017
```
cuducos: All these instructions contrast with lines 29 and 35 from `serenata_toolbox/serenata-toolbox`. There's no `federal-senate` or `reimbursements` to be entered in the CLI anymore.

willianpaixao: Whoops. Fixing.

PR #204, opened 2018-10-31
Fixing running serenata-toolbox into jupyter with tornado 5.0

thiagoalmeidasa:
This PR aims to fix running serenata-toolbox into jupyter with tornado 5.0.

Following the issue thread [here](https://github.com/jupyter/notebook/issues/3397), I changed the way to launch a task in the Downloader class.

Try to follow [this documentation step](https://github.com/okfn-brasil/serenata-toolbox#example-2-how-do-i-download-the-datasets) using a jupyter notebook with tornado 5.0 and face a `RuntimeError: This event loop is already running` exception. After that, try to install using my branch and realize that exception has disappeared.

Anyone involved in the project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,7 @@ jobs:
 
 install:
   - rm -rf .coverage
-  - python -m pip install -e .
-  - python -m pip install coveralls pytest pytest-cov
+  - python -m pip install -e .[dev]
 
 after_success:
   - coveralls
--- a/README.rst
+++ b/README.rst
@@ -141,7 +141,7 @@ Always add tests to your contribution — if you want to test it locally before
 
 .. code-block:: bash
 
-  $ pip install pytest pytest-cov
+  $ pip install -e .[dev]
   $ pytest
 
 When the tests are passing, also check for coverage of the modules you edited or added — if you want to check it before opening the PR:
@@ -155,14 +155,14 @@ Follow `PEP8 <https://www.python.org/dev/peps/pep-0008/>`_ and best practices im
 
 .. code-block:: bash
 
-  $ pip install prospector
+  $ pip install -e .[dev]
   $ prospector -s veryhigh serenata_toolbox
 
 If this report includes issues related to `import` section of your files, `isort <https://github.com/timothycrosley/isort>`_ can help you:
 
 .. code-block:: bash
 
-  $ pip install isort
+  $ pip install -e .[dev]
   $ isort **/*.py --diff
 
 Always suggest a version bump. We use `Semantic Versioning <http://semver.org>`_ – or in `Elm community words <https://github.com/elm-lang/elm-package#version-rules>`_:
--- a/serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py
+++ b/serenata_toolbox/chamber_of_deputies/reimbursements_cleaner.py
@@ -141,8 +141,9 @@ class ReimbursementsCleaner:
     def aggregate_multiple_payments(self):
         self.data = pd.concat([
             self._house_payments(),
-            self._non_house_payments(),
-        ])
+            self._non_house_payments()],
+            sort=True,
+        )
 
     def save(self):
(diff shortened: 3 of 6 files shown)


cuducos:
> This PR aims to fix running serenata-toolbox into jupyter with tornado 5.0.

Sorry @thiagoalmeidasa, I'm afraid I missed the point here — there's too much information packed in a simple sentence.

Would you mind expanding on:

* What is the context and use case of running the toolbox in jupyter with tornado 5?
* What errors are raised in this scenario?
* Will `create_task` work on different setups?

thiagoalmeidasa:
Thanks for the review @cuducos .

Let me try to explain.

> Would you mind expanding on:
> 
> * What is the context and use case of running the toolbox in jupyter with tornado 5?
> * What errors are raised in this scenario?

Some days ago I just tried to run [this documentation step](https://github.com/okfn-brasil/serenata-toolbox#example-2-how-do-i-download-the-datasets) using a fresh jupyter notebook installation and get this error: `RuntimeError: This event loop is already running`.

After that, I started to dig about this problem and found this [issue thread](https://github.com/jupyter/notebook/issues/3397), then tried to fix by reading and evaluate the suggestions proposed in the comments.

> * Will `create_task` work on different setups?

I'm not an expert in python development, but [this documentation section](https://docs.python.org/3/library/asyncio-task.html#awaitables) indicates that `create_task` will work fine in different setups.

cuducos:
My point is: I'm afraid that `create_task` would not invoke the coroutine in setups in which the loop is not already running (that is to say, outside tornado 5.0). Check this simple example I just coded:

```python
In [1]: import asyncio

In [2]: async def magic():
   ...:     print(42)
   ...:

In [3]: loop = asyncio.get_event_loop()

In [4]: loop.create_task(magic())
Out[4]: <Task pending coro=<magic() running at <ipython-input-2-ca8ce5b12955>:1>>
```

And compare to this one:

```python
In [1]: import asyncio

In [2]: async def magic():
   ...:     print(42)
   ...:

In [3]: loop = asyncio.get_event_loop()

In [4]: loop.run_until_complete(magic())
42
```

Maybe we need a way to know _if_ the loop is already running and use `create_task` or `run_until_complete` accordingly… what do you think?

thiagoalmeidasa:
Good point.

I'll try to improve this solution. Maybe @willianpaixao could help me to fix that.

willianpaixao:
I see a couple of ways to solve this problem:
1. You add an exception that is thrown in case an event loop already exists. The try/catch block should create a new thread-safe event loop. [[1]](https://docs.python.org/3/library/asyncio-task.html#scheduling-from-other-threads) [[2]](https://docs.python.org/3/library/asyncio-dev.html#asyncio-multithreading)
2. Before creating a task, check whether there is an existing event loop with `asyncio.get_running_loop()` and if so, append your task to it, otherwise, create a new event loop as usual with `asyncio.new_event_loop()` or `asyncio.get_event_loop()`. [[3]](https://docs.python.org/3/library/asyncio-eventloop.html)

cuducos:
As `get_running_loop`[ _can only be called from a coroutine or a callback_](
https://docs.python.org/3/library/asyncio-eventloop.htm)l, I think @willianpaixao's first options will result in a more readable code ; )

thiagoalmeidasa:
And what about that approach?

```
import asyncio

async def magic():
    print(42)


loop = asyncio.get_event_loop()

if loop.is_running():
    loop.create_task(magic())
else:
    loop.run_until_complete(magic())
```

It worked fine when using or not jupyter notebooks.

cuducos:
It looks great : )

cuducos:
I think we need two tests.

1. Mock the `loop.is_running()` so it returns `True` and we make two assertions: `loop.run_until.complete` should not be called and `loop.create_task` should be called
2. 1. Mock the `loop.is_running()` so it returns `False` and we make two assertions: `loop.run_until.complete` should be called and `loop.create_task` should not be called

PR #209, opened 2019-01-27
Danizavtz bug426

danizavtz:
https://github.com/okfn-brasil/serenata-de-amor/issues/426

This code comes from serenata-de-amor project, there is a bug addressed that this code should be merged in this repository.

Just copy pasted the folder research from serenata-de-amor into this repository

__Write how could it be tested and checked here__

@cuducos @jtemporal @willianpaixao

### DIFF
--- a/serenata_toolbox/research/requirements.txt
+++ b/serenata_toolbox/research/requirements.txt
@@ -1,0 +1,12 @@
+aiofiles==0.4.0
+aiohttp==3.5.1
+beautifulsoup4==4.7.0
+geopy==1.18.1
+grequests==0.3.0
+humanize==0.5.1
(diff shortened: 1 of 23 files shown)


cuducos:
I'm sorry for the lack of clarity, @danizavtz. But the idea was to move the functionality, not the code itself. What I mean is that the toolbox is:

* a `pip` installable package, so it would make no sense to have code note useful for this scenario in the repository
* also as a `pip` installable package, we would not include undocumented and untested code
* a package with an app architecture, so the new functionalities should fit this architecture

I'm closing this pull request because even if you wanna follow up on this pursuit, it worth it not to have this whole directory in the Git repository of this repo, ok?

PR #217, opened 2019-11-01
Support to Python 3.7 & Python 3.8

cuducos:

This is an updated version of #214 including Python 3.8.

Similar strategy of #214, but also updated the Docker image.

I used `tox`  and some manual tests.

@sergiomario 

**TODO**

I'm still not convinced Travis CI will work, so let's wait for the green lights!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git/
+.tox/
+data/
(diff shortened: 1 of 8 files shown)


cuducos:
Hum… CI is still failing (as I mentioned in the _TODO_). Does it worth it to open an issue ao we can publicize and look for help from Travis experts? (also the new version is not published at PyPI because it's Travis who does that)

lipemorais:
We need an issue to discus about it.

